PR #13, opened 2022-10-27
Add support for boot_volume_{size,type}

### LINKED ISSUE #5: Add support for `boot_volume_{size,type}`
There are a few labels that are used inside Magnum when doing boot from volume:

- `boot_volume_size`
- `boot_volume_type`

There is also corresponding set of functions that allow you to get the volume type:

https://github.com/openstack/magnum/blob/16bdedcf2fe6986c995bd415f4e3c70dac914ada/magnum/common/cinder.py#L25-L37

and some defaults:

- `default_boot_volume_size`: https://github.com/openstack/magnum/blob/5f9bfe3f3ca08776956866426351e385156ff37e/magnum/conf/cinder.py#L47-L50

It looks like for now, we'll be able to support boot from volume here:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/docs/book/src/clusteropenstack/configuration.md#boot-from-volume

so if `boot_volume_size` and `boot_volume_type` are set (or their defaults are), then we point to that.

okozachenko1203:
Fix #5

### DIFF
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -380,7 +380,7 @@ class OpenStackMachineTemplate(NodeGroupBase):
             },
             "imageUUID": self.node_group.image_id,
         }
-        boot_volume_size = get_label_value(
+        boot_volume_size = get_int_label_value(
             self.cluster,
             "boot_volume_size",
             CONF.cinder.default_boot_volume_size,
@@ -768,6 +768,14 @@ class Cluster(ClusterBase):
 def get_bool_label_value(cluster: any, key: str, default: bool) -> bool:
     value = get_label_value(cluster, key, default)
     return strutils.bool_from_string(value, strict=True)
+
+
+def get_int_label_value(cluster: any, key: str, default: bool) -> bool:
+    value = get_label_value(cluster, key, default)
+    if strutils.is_int_like(value):
+        return int(value)
+    msg = "%s is not an int-like string." % key
+    raise ValueError(msg)
 
 
 def get_label_value(cluster: any, key: str, default: str) -> str:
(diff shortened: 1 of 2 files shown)


okozachenko1203:
test

mnaser (changes requested):
@okozachenko1203 I think this breaks the use case if no volume size is set/enabled. 

Would it make more sense to instead have an if statement like I did for audit ?

mnaser:
```
Oct 27 13:52:48 devstack-magnum magnum-conductor[1094359]: ERROR oslo_messaging.rpc.server     if boot_volume_size > 0:
Oct 27 13:52:48 devstack-magnum magnum-conductor[1094359]: ERROR oslo_messaging.rpc.server TypeError: '>' not supported between instances of 'str' and 'int'
```

magnum_cluster_api/resources.py
```diff
@@ -378,7 +380,20 @@ def get_object(self) -> objects.OpenStackMachineTemplate:
             },
             "imageUUID": self.node_group.image_id,
         }
-
+        boot_volume_size = get_label_value(
```
mnaser, line 383: let's refactor this to get_int_label_value which parses the value into an in (similar to `get_bool_label_value`)

mnaser (changes requested):

magnum_cluster_api/resources.py
```diff
@@ -378,7 +380,20 @@ def get_object(self) -> objects.OpenStackMachineTemplate:
...
+            "boot_volume_size",
+            CONF.cinder.default_boot_volume_size,
+        )
+        if boot_volume_size > 0:
+            spec["rootVolume"] = {
+                "diskSize": boot_volume_size,
+                "volumeType": get_label_value(
+                    self.cluster,
+                    "boot_volume_type",
+                    cinder.get_default_boot_volume_type(self.context),
```
mnaser, line 394:
```
AttributeError: 'OpenStackMachineTemplate' object has no attribute 'context'
```

mnaser (changes requested):
I think it would be really good if you tested this in your environment.

mnaser (approved):

PR #19, opened 2022-11-07
Dynamically add all StorageClasses

### LINKED ISSUE #12: Dynamically add all StorageClass
An OpenStack cloud can have many volume types.  Instead of having a disconnect between the Kubernetes cluster and the OpenStack cloud, we should automatically poll for all the volume types and create `StorageClass` for them.

In the same time, we will also make the default `StorageClass` the one which is default in the cloud, that will make things nice and simple for the user.

okozachenko1203:
Fix #12

### DIFF
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -14,7 +14,7 @@ from oslo_config import cfg
 from oslo_serialization import base64
 from oslo_utils import encodeutils
 
-from magnum_cluster_api import objects, utils
+from magnum_cluster_api import clients, objects, utils
 
 CONF = cfg.CONF
 KUBE_TAG = "v1.25.3"
@@ -263,6 +263,68 @@ class CinderCSIClusterResourceSet(ClusterBase):
                     "resources": [
                         {
                             "name": f"cinder-csi-{version}",
+                            "kind": "ConfigMap",
+                        },
+                    ],
+                },
+            },
+        )
+
+
+class StorageClassesConfigMap(ClusterBase):
+    def get_object(self) -> pykube.ConfigMap:
+        osc = clients.get_openstack_api(context)
+        volume_types = osc.cinder().volume_types.list()
+        data = {}
+        for vt in volume_types:
+            data["%s.yaml" % vt.name] = yaml.dump(
+                {
+                    "apiVersion": objects.StorageClass.version,
+                    "kind": objects.StorageClass.kind,
+                    "metadata": {
+                        "annotations": {
+                            "storageclass.kubernetes.io/is-default-class": "true"
+                        } if vt.name == "__DEFAULT__" else {},
+                        "name": vt.name,
+                    },
+                    "provisioner": "kubernetes.io/cinder",
+                    "parameters": {
+                        "type": vt.name,
+                    },
+                    "reclaimPolicy": "Delete",
+                    "volumeBindingMode": "Immediate"
+                }
+            )
+        return pykube.ConfigMap(
+            self.api,
+            {
+                "apiVersion": pykube.ConfigMap.version,
+                "kind": pykube.ConfigMap.kind,
+                "metadata": {
+                    "name": f"storageclasses",
+                    "namespace": "magnum-system",
+                },
+                "data": data,
+            },
+        )
+
+
+class StorageClassesClusterResourceSet(ClusterBase):
+    def get_object(self) -> objects.ClusterResourceSet:
+
+        return objects.ClusterResourceSet(
+            self.api,
+            {
+                "apiVersion": objects.ClusterResourceSet.version,
+                "kind": objects.ClusterResourceSet.kind,
+                "metadata": {
+                    "name": f"storageclasses",
+                    "namespace": "magnum-system",
+                },
+                "spec": {
+                    "resources": [
+                        {
+                            "name": f"storageclasses",
                             "kind": "ConfigMap",
                         },
                     ],
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -39,6 +39,9 @@ class BaseDriver(driver.Driver):
         resources.CinderCSIConfigMap(self.k8s_api, cluster).apply()
         resources.CinderCSIClusterResourceSet(self.k8s_api, cluster).apply()
 
+        resources.StorageClassesConfigMap(self.k8s_api, cluster).apply()
+        resources.StorageClassesClusterResourceSet(self.k8s_api, cluster).apply()
+
         credential = osc.keystone().client.application_credentials.create(
             user=cluster.user_id,
             name=cluster.uuid,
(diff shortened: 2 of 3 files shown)


okozachenko1203:
ClusterResourceSet created correctly
```
apiVersion: v1
data:
  __DEFAULT__.yaml: |
    apiVersion: storage.k8s.io/v1
    kind: StorageClass
    metadata:
      annotations:
        storageclass.kubernetes.io/is-default-class: 'true'
      name: __DEFAULT__
    parameters:
      type: __DEFAULT__
    provisioner: kubernetes.io/cinder
    reclaimPolicy: Delete
    volumeBindingMode: Immediate
  lvmdriver-1.yaml: |
    apiVersion: storage.k8s.io/v1
    kind: StorageClass
    metadata:
      annotations: {}
      name: lvmdriver-1
    parameters:
      type: lvmdriver-1
    provisioner: kubernetes.io/cinder
    reclaimPolicy: Delete
    volumeBindingMode: Immediate
kind: ConfigMap
metadata:
  creationTimestamp: "2022-11-07T13:38:38Z"
  name: storageclasses
  namespace: magnum-system
  resourceVersion: "3002298"
  uid: 03f51036-61cc-43a6-8476-c926c2b5e526
```
```
apiVersion: addons.cluster.x-k8s.io/v1beta1
kind: ClusterResourceSet
metadata:
  creationTimestamp: "2022-11-07T13:44:16Z"
  finalizers:
  - addons.cluster.x-k8s.io
  generation: 1
  name: storageclasses
  namespace: magnum-system
  resourceVersion: "3003491"
  uid: 6ad15f32-a3b2-47f5-8585-a6513c4a6bb2
spec:
  clusterSelector:
    matchLabels:
      csi: cinder-csi-v1.25.3
  resources:
  - kind: ConfigMap
    name: storageclasses
  strategy: ApplyOnce
status:
  observedGeneration: 1
```

magnum_cluster_api/resources.py
```diff
@@ -271,6 +271,88 @@ def get_object(self) -> objects.ClusterResourceSet:
...
+            self.cluster, "cinder_csi_plugin_tag", CSI_TAG
+        )
+
+        return objects.ClusterResourceSet(
+            self.api,
+            {
+                "apiVersion": objects.ClusterResourceSet.version,
+                "kind": objects.ClusterResourceSet.kind,
+                "metadata": {
+                    "name": "storageclasses",
```
mnaser, line 336:
```suggestion
                    "name": "cinder-csi-storageclass",
```

I'm thinking we should do this to avoid clashes

magnum_cluster_api/resources.py
```diff
@@ -271,6 +271,88 @@ def get_object(self) -> objects.ClusterResourceSet:
...
+                "data": data,
+            },
+        )
+
+
+class StorageClassesClusterResourceSet(ClusterBase):
+    def get_object(self) -> objects.ClusterResourceSet:
+        version = utils.get_cluster_label(
+            self.cluster, "cinder_csi_plugin_tag", CSI_TAG
+        )
```
mnaser, line 328: I think we can drop this at this point, since we technically don't need this.

magnum_cluster_api/resources.py
```diff
@@ -271,6 +271,88 @@ def get_object(self) -> objects.ClusterResourceSet:
...
+                "apiVersion": objects.ClusterResourceSet.version,
+                "kind": objects.ClusterResourceSet.kind,
+                "metadata": {
+                    "name": "storageclasses",
+                    "namespace": "magnum-system",
+                },
+                "spec": {
+                    "clusterSelector": {
+                        "matchLabels": {
+                            "csi": f"cinder-csi-{version}",
```
mnaser, line 342:
```suggestion
                            "csi": "cinder",
```

We should swap to using this, so we match anything that has Cinder as a CSI

magnum_cluster_api/resources.py
```diff
@@ -271,6 +271,88 @@ def get_object(self) -> objects.ClusterResourceSet:
...
+                    "volumeBindingMode": "Immediate",
+                }
+            )
+        return pykube.ConfigMap(
+            self.api,
+            {
+                "apiVersion": pykube.ConfigMap.version,
+                "kind": pykube.ConfigMap.kind,
+                "metadata": {
+                    "name": "storageclasses",
```
mnaser, line 316:
```suggestion
                    "name": "cinder-csi-storageclass",
```

magnum_cluster_api/resources.py
```diff
@@ -271,6 +271,88 @@ def get_object(self) -> objects.ClusterResourceSet:
...
+                },
+                "spec": {
+                    "clusterSelector": {
+                        "matchLabels": {
+                            "csi": f"cinder-csi-{version}",
+                        },
+                    },
+                    "resources": [
+                        {
+                            "name": "storageclasses",
```
mnaser, line 347:
```suggestion
                            "name": "cinder-csi-storageclass",
```

magnum_cluster_api/resources.py
```diff
@@ -271,6 +271,88 @@ def get_object(self) -> objects.ClusterResourceSet:
...
+                    "apiVersion": objects.StorageClass.version,
+                    "kind": objects.StorageClass.kind,
+                    "metadata": {
+                        "annotations": {
+                            "storageclass.kubernetes.io/is-default-class": "true"
+                        }
+                        if vt.name == "__DEFAULT__"
+                        else {},
+                        "name": vt.name,
+                    },
```
mnaser, line 301: hmm, it's possible that the default storage class could be called something else other than `__DEFAULT__`

https://docs.openstack.org/api-ref/block-storage/v3/?expanded=list-all-volume-types-detail,show-default-volume-type-detail#show-default-volume-type

it's advised not to use/rely on `__DEFAULT__` so use that instead.

mnaser (changes requested):
https://github.com/vexxhost/magnum-cluster-api/pull/19/files#diff-43c5da84e3410f51531d7b4f6bdfe0b0f83fe9a19018cf1dde7ee2ef352bb019L1081-L1085

this part we should change to like

```
csi: cinder
cinder-csi-version: {{ version }}
```

and change this:

https://github.com/vexxhost/magnum-cluster-api/pull/19/files#diff-43c5da84e3410f51531d7b4f6bdfe0b0f83fe9a19018cf1dde7ee2ef352bb019R260

to be:

```
csi: cinder
cinder-csi-version: <version>
```

magnum_cluster_api/resources.py
okozachenko1203: ok, this is the exact method
https://github.com/openstack/python-cinderclient/blob/master/cinderclient/v3/volume_types.py#L116-L121

magnum_cluster_api/resources.py
```diff
@@ -295,7 +297,7 @@ def get_object(self) -> pykube.ConfigMap:
                         "annotations": {
                             "storageclass.kubernetes.io/is-default-class": "true"
                         }
-                        if vt.name == "__DEFAULT__"
+                        if default_volume_type and default_volume_type.name == vt.name
```
mnaser, line 300: is it possible that `default_volume_type` equals `None` ?

mnaser (changes requested):
small question :)

magnum_cluster_api/resources.py
okozachenko1203: The default_volume_type configuration option is required to have a value from the train version.
So this condition is meaningless.

mnaser (approved):

Issue #3, opened 2022-10-26
Autoscaling


mnaser:
Auto-scaling is a useful feature that used to exist inside Magnum, however, the Cluster API has [built-in autoscaling](https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/autoscaling.html).  We should enable it when `auto_scaling_enabled` is set to true.

I have not done significant research into this, so I will update this later once we have an idea how to get it all to work.

mnaser:
we'll be needing to deal/use this for autoscale:

https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/autoscaling.html#autoscaler-running-in-management-cluster-using-service-account-credentials-with-separate-workload-cluster

okozachenko1203:
on it

PR #39, opened 2022-11-11
Support autoscaling

okozachenko1203:
Fix https://github.com/vexxhost/magnum-cluster-api/issues/3


mnaser:
I believe in this case you've actually ran the auto scaler on the workload cluster with a kubeconfig file to talk to the management cluster

even if things can be secure and locked down, this could be a security issue down the road. Therefore, I think it's better to run the auto scaler on the management cluster talking to the workload instead. 

So I think you'll have to remove it from the ClusterResourceSet and get rid of the kubeconfig option, for the actual deployment in the management cluster, you can mount the clustername-kubeconfig secret to get access to the cluster directly.

okozachenko1203:
> I believe in this case you've actually ran the auto scaler on the workload cluster with a kubeconfig file to talk to the management cluster
> 
> even if things can be secure and locked down, this could be a security issue down the road. Therefore, I think it's better to run the auto scaler on the management cluster talking to the workload instead.
> 
> So I think you'll have to remove it from the ClusterResourceSet and get rid of the kubeconfig option, for the actual deployment in the management cluster, you can mount the clustername-kubeconfig secret to get access to the cluster directly.

when we run autoscaler in mgmt cluster, i wonder it means we need to deploy one autoscaler per one magnum cluster
That was my concern and the reason why i chose autosacler in workload cluster and tried to create clusterrole with resourceName binding.

mnaser:
> > I believe in this case you've actually ran the auto scaler on the workload cluster with a kubeconfig file to talk to the management cluster
> > even if things can be secure and locked down, this could be a security issue down the road. Therefore, I think it's better to run the auto scaler on the management cluster talking to the workload instead.
> > So I think you'll have to remove it from the ClusterResourceSet and get rid of the kubeconfig option, for the actual deployment in the management cluster, you can mount the clustername-kubeconfig secret to get access to the cluster directly.
> 
> when we run autoscaler in mgmt cluster, i wonder it means we need to deploy one autoscaler per one magnum cluster That was my concern and the reason why i chose autosacler in workload cluster and tried to create clusterrole with resourceName binding.

yes, indeed, actually we will need to create a autoscaler per CAPI cluster on the control plane, but I think the load added of this is kinda negligible so we can be OK with it.

okozachenko1203:
Ok, I have changed the clusterAPIMode as you mentioned.
Also I used flux helmrelease for autoscaler deployment.

magnum_cluster_api/driver.py
```diff
@@ -186,6 +192,10 @@ def delete_cluster(self, context, cluster):
         resources.ClusterResourceSet(self.k8s_api, cluster).delete()
         resources.ClusterResourcesConfigMap(context, self.k8s_api, cluster).delete()
         resources.Cluster(context, self.k8s_api, cluster).delete()
+        if utils.get_cluster_label_as_bool(self.cluster, "auto_scaling_enabled", False):
+            resources.ClusterAutoscalerHelmRelease(self.k8s_api, cluster).delete()
+            resources.ClusterAutoscalerHelmRepository(self.k8s_api, cluster).delete()
```
mnaser, line 197: not a good idea, HelmRepository is used by all of the clusters, so it should always be there :)

magnum_cluster_api/resources.py
```diff
@@ -85,6 +86,59 @@ def labels(self) -> dict:
...
+class ClusterAutoscalerHelmRelease(ClusterBase):
+    def get_object(self) -> objects.HelmRelease:
+        cluster_name = utils.get_or_generate_cluster_api_name(self.api, self.cluster)
+        return objects.HelmRelease(
+            self.api,
+            {
+                "apiVersion": objects.HelmRelease.version,
+                "kind": objects.HelmRelease.kind,
+                "metadata": {
+                    "name": self.cluster.uuid,
```
mnaser, line 117: let's use `cluster_name` here that way the deployment will be easier t see

mnaser (changes requested):
@okozachenko1203 good progress

magnum_cluster_api/driver.py
```diff
@@ -22,6 +22,9 @@
 class BaseDriver(driver.Driver):
     def __init__(self):
         self.k8s_api = clients.get_pykube_api()
+        resources.ClusterAutoscalerHelmRepository(
+            self.k8s_api,
+        ).apply()
```
mnaser, line 27: lets move this into `resources.apply_cluster_from_magnum_cluster`.

magnum_cluster_api/driver.py
```diff
@@ -186,6 +192,8 @@ def delete_cluster(self, context, cluster):
         resources.ClusterResourceSet(self.k8s_api, cluster).delete()
         resources.ClusterResourcesConfigMap(context, self.k8s_api, cluster).delete()
         resources.Cluster(context, self.k8s_api, cluster).delete()
+        if utils.get_cluster_label_as_bool(cluster, "auto_scaling_enabled", False):
+            resources.ClusterAutoscalerHelmRelease(self.k8s_api, cluster).delete()
```
mnaser, line 196: no need for the if, delete() will be noop if it exists

magnum_cluster_api/resources.py
```diff
@@ -85,6 +86,63 @@ def labels(self) -> dict:
...
+class ClusterAutoscalerHelmRelease(ClusterBase):
+    def get_object(self) -> objects.HelmRelease:
+        cluster_name = utils.get_or_generate_cluster_api_name(self.api, self.cluster)
+        return objects.HelmRelease(
+            self.api,
+            {
+                "apiVersion": objects.HelmRelease.version,
+                "kind": objects.HelmRelease.kind,
+                "metadata": {
+                    "name": f"autoscaler-{self.cluster.uuid}",
```
mnaser, line 117: lets use `cluster_name`

magnum_cluster_api/resources.py
```diff
@@ -1120,7 +1190,25 @@ def get_object(self) -> objects.Cluster:
...
-                                    "replicas": ng.node_count,
+                                    "replicas": None
+                                    if auto_scaling_enabled
+                                    else ng.node_count,
+                                    "metadata": {
+                                        "annotations": {
+                                            "cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size": utils.get_cluster_label(  # noqa: E501
+                                                self.cluster,
+                                                "min_node_count",
+                                                "0",
```
mnaser, line 1201: needs to be 1 at minimum

mnaser (changes requested):

magnum_cluster_api/resources.py
okozachenko1203: `cluster_name` is not proper because helmrelease name is used for k8s resource name which requires `DNS RFC1123`. But openstack resource name convention is not following the same rule as k8.
That is why I added `autoscaler-` as the prefix.

magnum_cluster_api/resources.py
```diff
@@ -85,6 +86,59 @@ def labels(self) -> dict:
...
+                                "kind": objects.HelmRepository.kind,
+                                "name": "autoscaler",
+                            },
+                        },
+                    },
+                    "values": {
+                        "cloudProvider": "clusterapi",
+                        "clusterAPIMode": "kubeconfig-incluster",
+                        "clusterAPIKubeconfigSecret": f"{cluster_name}-kubeconfig",
+                    },
```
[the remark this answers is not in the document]
mnaser: this is still pending

mnaser (changes requested):

magnum_cluster_api/resources.py
okozachenko1203: openstack resource name can be started with number.
But it is not allowed in k8s resource name.

okozachenko1203: aha, ok i found this
magnum resource naming rule is a bit different than others like network
`^[a-zA-Z][a-zA-Z0-9_.-]*$`

okozachenko1203: but k8s service name follows dns-1035
```
The Service "k8s-v1.25.3" is invalid: metadata.name: Invalid value: "k8s-v1.25.3": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

okozachenko1203: DNS 1035
```
contain at most 63 characters
contain only lowercase alphanumeric characters or '-'
start with an alphabetic character
end with an alphanumeric character
```

mnaser: @okozachenko1203 can you then make a change to the function that generate cluster name to replace dots with `-` ?

magnum_cluster_api/driver.py
```diff
@@ -45,6 +45,9 @@ def create_cluster(self, context, cluster, cluster_create_timeout):
             self.k8s_api, cluster
         ).apply()
 
+        if utils.get_cluster_label_as_bool(cluster, "auto_scaling_enabled", False):
+            resources.ClusterAutoscalerHelmRelease(self.k8s_api, cluster).apply()
```
mnaser, line 49: let's move this to `resources.apply_cluster_from_magnum_cluster`, after creating the `HelmRepository`.

mnaser (changes requested):

magnum_cluster_api/resources.py
okozachenko1203: let me try

PR #41, opened 2022-11-15
Override delete method of ClusterResourcesConfigMap

okozachenko1203:
Now get_object method has an assert. Once this assertion is negative, deletion is failed as well as creation. By overriding the deletion method using the config map name directly, it avoids deletion failure. Also, it will reduce API requests.


mnaser (changes requested):
@okozachenko1203 

let's switch to logging a warning instead of an assert and then always hardcode calico, that's better

mnaser (approved):

Issue #42, opened 2022-11-16
Add support for Flatcar Linux


mnaser:
related:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1261
https://github.com/kubernetes-sigs/image-builder/issues/939

lukasmrtvy:
ping, its doable now :)

stefan-chivu:
Hi @all, I was wondering why the work for supporting Flatcar Linux has been halted. 

I would be interested to contribute and I'd like to know if the work done so far has any reason to be stopped or if it just happened to get to the back of the list of priorities.

mnaser:
@stefan-chivu seems to have followed up!

https://kubernetes.slack.com/archives/C05Q8TDTK6Z/p1695293755953929

Hoping to see the PR soon :D

mnaser:
closed by @stefan-chivu -- awesome work!

PR #44, opened 2022-11-17
feat: Support Flatcar OS and custom NTP servers

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/42

fix https://github.com/vexxhost/magnum-cluster-api/issues/45

MAGNUM-2

### DIFF
--- a/magnum_cluster_api/cmd/image_builder.py
+++ b/magnum_cluster_api/cmd/image_builder.py
@@ -95,6 +95,8 @@ def main(operating_system, version, image_builder_version):
         "kubernetes_deb_version": f"{version.replace('v', '')}-00",
         "kubernetes_semver": f"{version}",
         "kubernetes_series": f"{kubernetes_series}",
+        # https://github.com/flatcar/Flatcar/issues/823
+        "ansible_user_vars": "oem_id=openstack",
     }
     with tempfile.NamedTemporaryFile(suffix=".json") as fp:
         fp.write(json.dumps(customization).encode("utf-8"))
--- a/magnum_cluster_api/cmd/image_builder.py
+++ b/magnum_cluster_api/cmd/image_builder.py
@@ -68,6 +68,8 @@ def main(operating_system, version, image_builder_version):
 
     click.echo("- Download image-builder")
     path = f"{ib_path}.tar"
+    # TODO(Oleks): Need to use https://github.com/okozachenko1203/image-builder
+    #              until https://github.com/kubernetes-sigs/image-builder/pull/882 is fixed.
     with requests.get(
         f"https://github.com/kubernetes-sigs/image-builder/tarball/{image_builder_version}",
         stream=True,
--- a/magnum_cluster_api/cmd/image_builder.py
+++ b/magnum_cluster_api/cmd/image_builder.py
@@ -27,7 +27,7 @@ QEMU_PACKAGES = [
     "--operating-system",
     show_default=True,
     default="ubuntu-2004",
-    type=click.Choice(["ubuntu-2004"]),
+    type=click.Choice(["ubuntu-2004", "flatcar"]),
     help="Operating system to build image for",
 )
 @click.option(
--- a/magnum_cluster_api/cmd/image_builder.py
+++ b/magnum_cluster_api/cmd/image_builder.py
@@ -68,9 +68,6 @@ def main(operating_system, version, image_builder_version):
 
     click.echo("- Download image-builder")
     path = f"{ib_path}.tar"
-    # TODO(Oleks): Need to use https://github.com/okozachenko1203/image-builder
-    #              until https://github.com/kubernetes-sigs/image-builder/issues/1018 is fixed.
-    #              Or we can add those dropins into kubeadmConfigSpec ignition.
     with requests.get(
         f"https://github.com/kubernetes-sigs/image-builder/tarball/{image_builder_version}",
         stream=True,
--- a/magnum_cluster_api/cmd/image_builder.py
+++ b/magnum_cluster_api/cmd/image_builder.py
@@ -69,7 +69,8 @@ def main(operating_system, version, image_builder_version):
     click.echo("- Download image-builder")
     path = f"{ib_path}.tar"
     # TODO(Oleks): Need to use https://github.com/okozachenko1203/image-builder
-    #              until https://github.com/kubernetes-sigs/image-builder/pull/882 is fixed.
+    #              until https://github.com/kubernetes-sigs/image-builder/issues/1018 is fixed.
+    #              Or we can add those dropins into kubeadmConfigSpec ignition.
     with requests.get(
         f"https://github.com/kubernetes-sigs/image-builder/tarball/{image_builder_version}",
         stream=True,
--- a/magnum_cluster_api/cmd/image_builder.py
+++ b/magnum_cluster_api/cmd/image_builder.py
@@ -25,10 +25,9 @@ QEMU_PACKAGES = [
 @click.command()
 @click.option(
     "--operating-system",
-    show_default=True,
-    default="ubuntu-2004",
     type=click.Choice(["ubuntu-2004", "flatcar"]),
     help="Operating system to build image for",
+    prompt="Operating system to build image for",
 )
 @click.option(
     "--version",
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -917,6 +917,17 @@ class ClusterClass(Base):
                                     "jsonPatches": [
                                         {
                                             "op": "add",
+                                            "path": "/spec/template/spec/kubeadmConfigSpec/preKubeadmCommands",
+                                            "value": [
+                                                textwrap.dedent(
+                                                    """\
+                                                bash -c "sed -i 's/__REPLACE_NODE_NAME__/$(hostname -s)/g' /etc/kubeadm.yml"
+                                                """
+                                                )
+                                            ],
+                                        },
+                                        {
+                                            "op": "add",
                                             "path": "/spec/template/spec/kubeadmConfigSpec/format",
                                             "value": "ignition",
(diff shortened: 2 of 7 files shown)


mnaser:
by doing this, we're fully committing to Flatcar and not running anything on Ubuntu.

I think the cleaner thing is that we should create a variable called `operatingSystem` in the `ClusterClass`, and then in there, it can be an enum of either `ubuntu` or `flatcar`, by default it will be set to `ubuntu`.

And then, we should create a patch, conditional on `operatingSystem` being equal to `flatcar` to make all the changes that are needed for Flatcar, that way we can support both.

Then, we will have to update the driver to detect `os_distro` and use the appropriate model.  Also, I think that means we'll also need to update the image builder script to allow building flatcar images and publish them too.

okozachenko1203:
flatcar image here

https://object-storage.public.mtl1.vexxhost.net/swift/v1/7d9c8dc3734742b8a329c22b5c78bab3/magnum-capi-driver/flatcar-3374.2.0-v1.25.3

magnum_cluster_api/resources.py
```diff
@@ -419,6 +419,19 @@ def get_object(self) -> objects.KubeadmControlPlaneTemplate:
                     "template": {
                         "spec": {
                             "kubeadmConfigSpec": {
+                                "format": "ignition",
+                                "ignition": {
+                                    "containerLinuxConfig": {
+                                        "additionalConfig": "storage:\n  links:\n  - path: /etc/systemd/system/kubeadm.service.wants/containerd.service\n    target: /usr/lib/systemd/system/containerd.service\n"
```
mnaser, line 425: I think you can use `textwrap.dedent` and have a cleaner multiline string here :)

magnum_cluster_api/resources.py
```diff
@@ -419,6 +419,19 @@ def get_object(self) -> objects.KubeadmControlPlaneTemplate:
...
+                                    },
+                                },
+                                # In Flatcar kubeadm configuration is in different directory because /run
+                                # can't be provisioned with ignition.
+                                "preKubeadmCommands": [
+                                    """
+                                    bash -c "sed -i 's/__REPLACE_NODE_NAME__/$(hostname -s)/g' /etc/kubeadm.yml"
+                                    """
```
mnaser, line 433: why couldn't we use `{{ local_hostname }}` -- does that not work with ignition or?

mnaser (changes requested):
good progress, let's make sure to switch to patches on the ClusterClass as mentioned

Also, let's update title to be `feat` since this is a feature addition.

magnum_cluster_api/resources.py
okozachenko1203: `{{ local_hostname }}` is instance metadata from cloud-init.
Flatcar ignition doesn't support such metadatas.
Btw, Flatcar Container Linux ships with a small utility, coreos-metadata, which fetches information specific to the environment in which Flatcar Container Linux is running.  https://github.com/flatcar/afterburn/blob/master/docs/container-linux-legacy.md
I will use `${COREOS_OPENSTACK_HOSTNAME}` instead.

mnaser (changes requested):
small nits, otherwise is this ready to land @okozachenko1203 -- shall we drop the WIP?

magnum_cluster_api/cmd/image_builder.py
```diff
@@ -27,7 +27,7 @@
     "--operating-system",
     show_default=True,
     default="ubuntu-2004",
-    type=click.Choice(["ubuntu-2004"]),
+    type=click.Choice(["ubuntu-2004", "flatcar"]),
```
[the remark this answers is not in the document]
okozachenko1203: well, I tried that before but `os_distro` and operating system name are not matching exactly, for example, `ubuntu-2004` in image-builder while `ubuntu-focal` in openstack `os_distro`

okozachenko1203:
> > small nits, otherwise is this ready to land @okozachenko1203 -- shall we drop the WIP?
> 
> @mnaser All patches are rendered exactly but now I am facing kcp unavailable issue. Seems [kubernetes-sigs/image-builder#939 (comment)](https://github.com/kubernetes-sigs/image-builder/issues/939#issuecomment-1267318988) is not working. Btw, I need to dig more by jump into the machine. But key auth is not working with flatcar image. Created an issue to the upstream. [Flatcar images built by image-builder does not support key auth · Issue #1018 · kubernetes-sigs/image-builder](https://github.com/kubernetes-sigs/image-builder/issues/1018)


I close this https://github.com/kubernetes-sigs/image-builder/issues/1018 and land openstack provider specific override in this driver.

magnum_cluster_api/cmd/image_builder.py
```diff
@@ -96,6 +95,8 @@ def main(operating_system, version, image_builder_version):
         "kubernetes_deb_version": f"{version.replace('v', '')}-00",
         "kubernetes_semver": f"{version}",
         "kubernetes_series": f"{kubernetes_series}",
+        # https://github.com/flatcar/Flatcar/issues/823
+        "ansible_user_vars": "oem_id=openstack",
```
mnaser, line 99: nice catch here.

mnaser: There is a lot of hacks around this because we're using an ISO image, @okozachenko1203 is going to contribute to `image-builder` an `openstack` image so that it uses OpenStack to deploy and publish directly to Glance.

Once we have that in place, we should switch the image builder to use that instead.  It'll drive down the dependencies  and we can remove this after.

magnum_cluster_api/resources.py
```diff
@@ -891,6 +915,238 @@ def get_object(self) -> objects.ClusterClass:
...
+                                                                # Set Openstack as coreos-metadata provider
+                                                                Environment=COREOS_METADATA_OPT_PROVIDER=--provider=openstack-metadata
+                                                          - name: coreos-metadata-sshkeys@.service
+                                                            enabled: true
+                                                            dropins:
+                                                            - name: 20-clct-provider-override.conf
+                                                              contents: |
+                                                                [Service]
+                                                                # Set Openstack as coreos-metadata provider
+                                                                Environment=COREOS_METADATA_OPT_PROVIDER=--provider=openstack-metadata
```
mnaser, line 1033: we can remove this after moving to openstack built images.

magnum_cluster_api/resources.py
```diff
@@ -891,6 +915,238 @@ def get_object(self) -> objects.ClusterClass:
...
+                                                              contents: |
+                                                                [Service]
+                                                                Environment=COREOS_METADATA_OPT_PROVIDER=--provider=openstack-metadata
+                                                          - name: coreos-metadata-sshkeys@.service
+                                                            enabled: true
+                                                            dropins:
+                                                            - name: 20-clct-provider-override.conf
+                                                              contents: |
+                                                                [Service]
+                                                                Environment=COREOS_METADATA_OPT_PROVIDER=--provider=openstack-metadata
```
mnaser, line 1121: we can remove this after moving to openstack built images

magnum_cluster_api/resources.py
```diff
@@ -891,6 +915,238 @@ def get_object(self) -> objects.ClusterClass:
...
+                                                                [Unit]
+                                                                # kubeadm must run after coreos-metadata populated /run/metadata directory.
+                                                                Requires=coreos-metadata.service
+                                                                After=coreos-metadata.service
+                                                                [Service]
+                                                                # Ensure kubeadm service has access to kubeadm binary in /opt/bin on Flatcar.
+                                                                Environment=PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/opt/bin
+                                                                # To make metadata environment variables available for pre-kubeadm commands.
+                                                                EnvironmentFile=/run/metadata/*
+                                                        """  # noqa: E501
```
mnaser, line 1047: this feels like it should live inside `image-builder`, so I think as a separate PR we should fix this?

magnum_cluster_api/resources.py
```diff
@@ -891,6 +915,238 @@ def get_object(self) -> objects.ClusterClass:
...
+                                                                # kubeadm must run after coreos-metadata populated /run/metadata directory.
+                                                                Requires=coreos-metadata.service
+                                                                After=coreos-metadata.service
+                                                                [Service]
+                                                                # In Flatcar /usr is immutable, so image-builder puts the binaries in /opt/bin instead.
+                                                                # Ensure kubeadm service has access to kubeadm binary in /opt/bin on Flatcar.
+                                                                Environment=PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/opt/bin
+                                                                # To make metadata environment variables available for pre-kubeadm commands.
+                                                                EnvironmentFile=/run/metadata/*
+                                                        """  # noqa: E501
```
mnaser, line 1136: same for this

mnaser (changes requested):

magnum_cluster_api/resources.py
invidian: The PATH modifying part should not be needed anymore since https://github.com/kubernetes-sigs/image-builder/commit/f4864de5cad4a826aff80259bf6681e37a49f18d, but I think dependence on `coreos-metadata.service` is platform-specific, so I don't think `image-builder` is a good place to put this.

mnaser:
I think perhaps maybe if there's some documentation on building images and how clusters are defined for a flatcar system.. that would be helpful

Mazorius:
I am waiting for the NTP fix as on autoscaling it is not possible to always change the internal NTP server.

Can the NTP fix be extracted so it can be merged @okozachenko1203 ?

mnaser:
@Mazorius you shouldn't have any ntp issues with the latest images as you'll be getting the NTP server from DHCP, and the Flatcar part has been merged so I'm closing thsi

PR #59, opened 2023-01-17
feat: Use crane for image loader instead of skopeo

okozachenko1203:
In addition, disable MD012 (Multiple consecutive blank lines) mdl lint rule.
This is annoying us against CHANGELOG.md file which is generated by release-please process.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,9 +13,9 @@ click = "*"
 requests = "*"
 shortuuid = "*"
 certifi = "*"
-"oslo.log" = "^5.0.2"
-"oslo.config" = "^9.0.0"
-"oslo.concurrency" = "^5.0.1"
+"oslo.log" = "*"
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -13,6 +13,9 @@ click = "*"
 requests = "*"
 shortuuid = "*"
 certifi = "*"
+"oslo.log" = "^5.0.2"
+"oslo.config" = "^9.0.0"
+"oslo.concurrency" = "^5.0.1"
```
mnaser, line 18: it is possible that magnum-capi will get installed in many different versions of OpenStack, which will have different locks on this.

Can you move this to `*` so that it will work as long as the dependency is there.

I doubt the API of those will be changing anytime soon.

mnaser (changes requested):

Issue #70, opened 2023-02-16
routing issues due to podSubnet and serviceSubnet overlapping


robincron:
executing `k get cm kubeadm-config -n kube-system -oyaml `in a cluster shows that podSubnet and serviceSubnet overlap, which could cause routing issues (also see https://github.com/coredns/coredns/issues/3704)

mnaser:
Fixed by #71

PR #71, opened 2023-02-16
fix: add label calico_ipv4_pool

robincron:
fix for issue https://github.com/vexxhost/magnum-cluster-api/issues/70

### DIFF
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -38,7 +38,6 @@ CLOUD_PROVIDER_TAG = "v1.25.3"
 CALICO_TAG = "v3.24.2"
 AUTOSCALER_HELM_CHART_VERSION = "9.21.0"
 CSI_TAG = "v1.25.3"
-CALICO_IPV4_POOL = "10.200.0.0/16"
 
 CLUSTER_CLASS_VERSION = pkg_resources.require("magnum_cluster_api")[0].version
 CLUSTER_CLASS_NAME = f"magnum-v{CLUSTER_CLASS_VERSION}"
@@ -1145,13 +1144,9 @@ class Cluster(ClusterBase):
             self.cluster, "cloud_provider_tag", CLOUD_PROVIDER_TAG
         )
         cni_version = utils.get_cluster_label(self.cluster, "calico_tag", CALICO_TAG)
-        calico_ipv4pool = utils.get_cluster_label(
-            self.cluster, "calico_ipv4pool", CALICO_IPV4_POOL
-        ) 
         labels = {
             "cni": f"calico-{cni_version}",
             "ccm": f"openstack-cloud-controller-manager-{ccm_version}",
-            "calico_ipv4pool": calico_ipv4pool,
         }
 
         if utils.get_cluster_label_as_bool(self.cluster, "cinder_csi_enabled", True):
@@ -1196,7 +1191,16 @@ class Cluster(ClusterBase):
                                 utils.get_cluster_label(
                                     self.cluster,
                                     "calico_ipv4pool",
-                                    "10.200.0.0/16",
+                                    "10.100.0.0/16",
+                                )
+                            ],
+                        },
+                        "services": {
+                            "cidrBlocks": [
+                                utils.get_cluster_label(
+                                    self.cluster,
+                                    "service_cluster_ip_range",
+                                    "10.254.0.0/16",
                                 )
                             ],
                         },
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -1151,7 +1151,7 @@ class Cluster(ClusterBase):
         labels = {
             "cni": f"calico-{cni_version}",
             "ccm": f"openstack-cloud-controller-manager-{ccm_version}",
-            "calico_ipv4pool": f"{calico_ipv4pool}",
+            "calico_ipv4pool": calico_ipv4pool,
         }
 
         if utils.get_cluster_label_as_bool(self.cluster, "cinder_csi_enabled", True):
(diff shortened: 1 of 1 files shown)


robincron:
cluster creation fails with error:
Cluster.cluster.x-k8s.io "test-rcr-afterpatch-7fvw3ma6en" is invalid: metadata.labels: Invalid value: "10.200.0.0/16": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?') |

mnaser:
@robincron so I think we don't want to change the default since we want to use the same old value as Magnum.

As you see from here:

https://opendev.org/openstack/magnum/src/branch/master/doc/source/user/index.rst#labels

The value used for `service_cluster_ip_range` is `10.254.0.0/16` -- so we should default to that

magnum_cluster_api/resources.py
```diff
@@ -38,6 +38,7 @@
 CALICO_TAG = "v3.24.2"
 AUTOSCALER_HELM_CHART_VERSION = "9.21.0"
 CSI_TAG = "v1.25.3"
+CALICO_IPV4_POOL = "10.200.0.0/16"
```
mnaser, line 41: we can skip this here and keep it hard coded below.

magnum_cluster_api/resources.py
```diff
@@ -1143,11 +1144,14 @@ def labels(self) -> dict:
         ccm_version = utils.get_cluster_label(
             self.cluster, "cloud_provider_tag", CLOUD_PROVIDER_TAG
         )
-        cni_verison = utils.get_cluster_label(self.cluster, "calico_tag", CALICO_TAG)
-
+        cni_version = utils.get_cluster_label(self.cluster, "calico_tag", CALICO_TAG)
+        calico_ipv4pool = utils.get_cluster_label(
+            self.cluster, "calico_ipv4pool", CALICO_IPV4_POOL
+        )
```
mnaser, line 1150: we can drop this, since it doesnt really affect the deployment of the 'cluster addon'

magnum_cluster_api/resources.py
```diff
@@ -1143,11 +1144,14 @@ def labels(self) -> dict:
...
-
+        cni_version = utils.get_cluster_label(self.cluster, "calico_tag", CALICO_TAG)
+        calico_ipv4pool = utils.get_cluster_label(
+            self.cluster, "calico_ipv4pool", CALICO_IPV4_POOL
+        ) 
         labels = {
-            "cni": f"calico-{cni_verison}",
+            "cni": f"calico-{cni_version}",
             "ccm": f"openstack-cloud-controller-manager-{ccm_version}",
+            "calico_ipv4pool": calico_ipv4pool,
```
mnaser, line 1154: and this too

magnum_cluster_api/resources.py
```diff
@@ -1192,7 +1196,7 @@ def get_object(self) -> objects.Cluster:
                                 utils.get_cluster_label(
                                     self.cluster,
                                     "calico_ipv4pool",
-                                    "10.100.0.0/16",
+                                    "10.200.0.0/16",
                                 )
                             ],
                         },
```
mnaser, line 1202: according to this:

https://cluster-api.sigs.k8s.io/tasks/bootstrap/kubeadm-bootstrap.html#kubeadmconfig-objects

we should add the following:

```python
"service": {
  "cidrBlocks": [
    utils.get_cluster_label(
      self.cluster,
      "service_cluster_ip_range",
      "10.254.0.0/16"
    )
  ]
}
```

mnaser (changes requested):

magnum_cluster_api/resources.py
robincron: removed it https://github.com/vexxhost/magnum-cluster-api/commit/84f1fb3d3870ed8be5f5f436dc01f7ce8ff69ad9

robincron:
added https://github.com/vexxhost/magnum-cluster-api/pull/71/commits/84f1fb3d3870ed8be5f5f436dc01f7ce8ff69ad9 according to https://github.com/vexxhost/magnum-cluster-api/pull/71#discussion_r1108607612 and https://github.com/kubernetes-sigs/cluster-api/blob/main/config/crd/bases/cluster.x-k8s.io_clusters.yaml

built image locally and tested, seems to do the job:

```
k get cm kubeadm-config -oyaml -n kube-system
apiVersion: v1
data:
  ClusterConfiguration: |
    apiServer:
      extraArgs:
        authorization-mode: Node,RBAC
        cloud-provider: external
        profiling: "false"
      timeoutForControlPlane: 4m0s
    apiVersion: kubeadm.k8s.io/v1beta3
    certificatesDir: /etc/kubernetes/pki
    clusterName: test-rcr-10-06-ggzsfdcx4a
    controlPlaneEndpoint: 10.6.12.233:6443
    controllerManager:
      extraArgs:
        cloud-provider: external
        profiling: "false"
    dns: {}
    etcd:
      local:
        dataDir: /var/lib/etcd
    imageRepository: container-infra-registry.fieldtest.c3.cgm.ag/magnum
    kind: ClusterConfiguration
    kubernetesVersion: v1.23.13
    networking:
      dnsDomain: cluster.local
      podSubnet: 10.100.0.0/16
      serviceSubnet: 10.254.0.0/16
    scheduler:
      extraArgs:
        profiling: "false"
kind: ConfigMap
metadata:
  creationTimestamp: "2023-02-17T09:10:51Z"
  name: kubeadm-config
  namespace: kube-system
  resourceVersion: "354"
  uid: 657d3e9d-6388-4b31-bd83-8f7b59ea79ee
```

mnaser (approved):

Issue #79, opened 2023-03-28
Add Manila CSI


mnaser:
We should detect what services are running in the cloud (such as Manila or Cinder) and then install the appropriate CSI.

We already deploy Cinder but all the time, instead, we should deploy it if we detect the Cinder service, and the same goes for Manila, that way we finally have `ReadWriteMany` volumes! :)

mnaser:
I think the direction we'll probably go in this is that the user will have to create a `StorageClass` on their own afterwards pointing at the correct `shareNetwork`.

okozachenko1203:
WIP: https://github.com/vexxhost/magnum-cluster-api/pull/80

okozachenko1203:
> I think the direction we'll probably go in this is that the user will have to create a `StorageClass` on their own afterwards pointing at the correct `shareNetwork`.

@mnaser  yeah, surely.
because a share_network is a tenant-defined object and customers can choose on-fly, we cannot detect it programmatically.

How about this? I think we can have one more label `manila_shared_network_id`, if it is not specified, we will skip creating the default SC(StorageClass) for manila and if it is specified, we will create a default SC using that.
Even if users didn't specify it as coe cluster label, they can create SC themselves after the cluster creation

okozachenko1203:
one more thing is `shareNetworkID` is only mandatory when `driver_handles_share_servers=true`. if users didn't enable it, we don't need to require share_network_id to create the RWM storageclass. 
So we have to detect/know `driver_handles_share_servers` value and decide the required params for SC creation.
Or we can just leave this kind of configuration for SC creation to the user instead of such a complex detection>

PR #80, opened 2023-03-28
feat: Add manila csi

okozachenko1203:
- This PR includes manila-csi-plugin deployment. This PR extends `magnum.conf` by adding `[manila_client]` section for keystone authtoken. So magnum driver can configure manila_client and use this to interact with sharev2 api.
- `magnum` doesn't support `manila` intergration so I override `magnum.common.clients.OpenStackClients` to include `manila` method which returns the manila client.
- This PR also makes cinder-csi-plugin and manila-csi-plugin as conditional.  i.e. even if labels (`cinder_csi_enabled`, `manila_csi_enabled`) are set as `true`, if the underlying openstack has no enabled endpoints for cinder and manila, it will not install those plugins.

fix: https://github.com/vexxhost/magnum-cluster-api/issues/79


magnum_cluster_api/utils.py
```diff
@@ -233,3 +236,56 @@ def delete_loadbalancers(ctx, cluster):
...
+        manila_svc = keystone.client.services.list(type="sharev2")
+    except Exception:
+        LOG.exception("Failed to list services")
+        raise exception.ServicesListFailed()
+
+    if manila_svc and manila_svc[0].enabled:
+        return True
+
+    LOG.info("There is no sharev2 service enabled in the cloud.")
+    return False
```
okozachenko1203, line 277: to detect services in the cloud

okozachenko1203 (commented):
@mnaser i added comments where requires more attention.

mnaser:
@okozachenko1203 I previously built a tool here which will 'sync' all the manifests and apply all the changes that we use -- https://github.com/vexxhost/magnum-cluster-api/blob/main/tools/sync-csi-manifests

There are other sync jobs too.  I think what I would like is that maybe we extend that script, and we also create a job that runs this script in CI and uses git to make sure that no changes happen, which means that no one made changes that were not implmented in this script and we can 'reliabely' get the manifests by running that script.

The potential second step to this is adding this script into the build script so that the manifestis not bundled in the source code but added in build time like we do for charts.

okozachenko1203:
> @okozachenko1203 I previously built a tool here which will 'sync' all the manifests and apply all the changes that we use -- https://github.com/vexxhost/magnum-cluster-api/blob/main/tools/sync-csi-manifests
> 
> There are other sync jobs too. I think what I would like is that maybe we extend that script, and we also create a job that runs this script in CI and uses git to make sure that no changes happen, which means that no one made changes that were not implmented in this script and we can 'reliabely' get the manifests by running that script.
> 
> The potential second step to this is adding this script into the build script so that the manifestis not bundled in the source code but added in build time like we do for charts.

I think we can consider keeping diff/patch files instead of writing py script.

PR #81, opened 2023-03-31
feat: Eliminate flux dependency for cluster autoscaler

okozachenko1203:


mnaser:
@okozachenko1203 did you forget `git add` by any chance? :)

okozachenko1203:
> @okozachenko1203 did you forget `git add` by any chance? :)

aha, my bad

magnum_cluster_api/resources.py
```diff
@@ -47,6 +49,99 @@
...
+
+
+class HelmRelease(BaseCli):
+    APPLY_CMD = "helm upgrade --install".split()
+    DELETE_CMD = "helm delete".split()
+    CHART_SOURCE = ""
+    RELEASE_NAME = ""
+
+    RELEASE_NAMESPACE = ""
+    EXTRA_ARGS = []
```
mnaser, line 91: lets drop this since its not being used

magnum_cluster_api/resources.py
```diff
@@ -47,6 +49,99 @@
...
+    VALUES = {}
+
+    def __init__(self) -> None:
+        assert self.RELEASE_NAME
+        assert self.CHART_SOURCE
+        self.APPLY_CMD += [
+            self.RELEASE_NAME,
+            self.CHART_SOURCE,
+            *self.EXTRA_ARGS,
```
mnaser, line 100: lets drop this since its not being used

magnum_cluster_api/resources.py
```diff
@@ -47,6 +49,99 @@
...
+    APPLY_CMD = "helm upgrade --install".split()
+    DELETE_CMD = "helm delete".split()
+    CHART_SOURCE = ""
+    RELEASE_NAME = ""
+
+    RELEASE_NAMESPACE = ""
+    EXTRA_ARGS = []
+    VALUES = {}
+
+    def __init__(self) -> None:
```
mnaser, line 94:
```suggestion
    def __init__(self, release_name: str, chart: str, namespace: typing.Optional[str]=None, values: dict) -> HelmRelease:
```

and then I suggest storing those local values as `self.` in lower case since they are not constants.

magnum_cluster_api/resources.py
```diff
@@ -47,6 +49,99 @@
...
+        assert self.RELEASE_NAME
+        assert self.CHART_SOURCE
+        self.APPLY_CMD += [
```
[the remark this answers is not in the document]
mnaser: also, in order for upgrade to be idempotent, we should use the following:

`helm upgrade --install --wait`

that way, if it already exists for some reason, it will just override.  also, it will make be helpful to wait and make sure it successfully goes up.

mnaser (changes requested):

magnum_cluster_api/resources.py
```diff
@@ -47,6 +49,99 @@
...
+
+    def __init__(self, api, cluster) -> None:
+        cluster_name = utils.get_or_generate_cluster_api_name(api, cluster)
+        self.RELEASE_NAME = cluster_name
+
+        manifests_path = pkg_resources.resource_filename(
+            "magnum_cluster_api", "manifests"
+        )
+        self.CHART_SOURCE = os.path.join(manifests_path, "cluster-autoscaler/")
+        self.VALUES = {
```
mnaser, line 130: we should override the image as well to accomodate environments that are air-gapped.

mnaser:
so the only thing left here is enabling the configuration of the `image_repository` for the autoscaler.

PR #114, opened 2023-05-22
fix: Wait until observedGeneration of the capi cluster is increased i…

### LINKED ISSUE #57: Cluster status is not updated properly
I've been noticing that on every action took on a magnum cluster, it delays up to 1 minute to have it reflected on magnum api. 

In some cases, like a sequence of rolling upgrades, the second time we upgrade a cluster ( with UPDATE_COMPLETE status ), it keeps doing the upgrade, but it doesnt update the magnum status to UPDATE_IN_PROGESS. 

This might confuse the end-user on what is the operation being done at the moment.

okozachenko1203:
…n cluster upgrade

fix #57

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -13,6 +13,7 @@
 # under the License.
 
 import time
+from tenacity import retry, stop_after_delay, wait_fixed
 
 import keystoneauth1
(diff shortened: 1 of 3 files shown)


mnaser:
Could you use "tenacity" which is also one of of our existing dependencies?

PR #121, opened 2023-06-07
chore: Add conditions to check if cluster.stack_id is set

okozachenko1203:


magnum_cluster_api/resources.py
```diff
@@ -1440,6 +1440,8 @@ def labels(self) -> dict:
         return {**super().labels, **labels}
 
     def get_or_none(self) -> objects.Cluster:
+        if self.cluster.stack_id is None:
+            return None
```
mnaser, line 1444: i think this is not necessary because we're doing `get_or_none` so if it cant find it, it will just return `None` anyways.

magnum_cluster_api/resources.py
```diff
@@ -1709,6 +1711,8 @@ def apply_cluster_from_magnum_cluster(
 def get_kubeadm_control_plane(
     api: pykube.HTTPClient, cluster: magnum_objects.Cluster
 ) -> objects.KubeadmControlPlane:
+    if cluster.stack_id is None:
+        return None
```
mnaser, line 1715: wouldn't the filter below return 0 results, and because it gave 0 results, it will not do the if statement and simply return None?

magnum_cluster_api/resources.py
```diff
@@ -1724,6 +1728,8 @@ def get_machine_deployment(
     cluster: magnum_objects.Cluster,
     node_group: magnum_objects.NodeGroup,
 ) -> objects.KubeadmControlPlane:
+    if cluster.stack_id is None:
+        return None
```
mnaser, line 1732: same thing here, it will return nothing since nothing is matching and since the if statement is not equal to 1 then return None?

okozachenko1203: yeap, i checked that `kubectl get deploy -l test1=null, test2=test2` doesn't return deployments which has only `test2=test2` label.

Issue #64, opened 2023-01-20
Bubble errors up to Magnum API


mnaser:
At the moment, errors in the Cluster API do not bubble up into the Magnum API

```
  Warning  Failedcreatenetwork  11m (x19 over 34m)  openstack-controller  Failed to create network k8s-clusterapi-cluster-magnum-system-cluster-2-ikhba5qojt: Expected HTTP response code [201 202] when accessing [POST https://network.openstack.cloud.local/v2.0/networks], but got 409 instead
{"NeutronError": {"type": "OverQuota", "message": "Quota exceeded for resources: ['network'].", "detail": ""}}
```

We need to bubble those errors up to OpenStack, or at least the quota checks.

fnpanic:
@mnaser from my point of view all errors should be reflected in magnum api with a reason so the user has a chance to understand (in terms of a quota issue) or addressed with the operators in a straight forward way.

mnaser:
> @mnaser from my point of view all errors should be reflected in magnum api with a reason so the user has a chance to understand (in terms of a quota issue) or addressed with the operators in a straight forward way.

Indeed, but there's a really weird scenario where the CAPI is like a controller so it will keep retrying, so it's not like it's technically "CREATE_FAILED" and the process is over.. technically if you fix the quota, the process will continue... so maybe "CREATE_BLOCKED" with a message might make sense, but that is not a native Magnum API concept.

fnpanic:
That is a good to know because i am pretty sure we will have users which keep clusters in failed stated forever which means the controller will retry forever which will place some unnessecary load on all openstack APIs. I like the idea of CREATE_BLOCKED but not sure if it is really a good idea to extend it. If we will not add changes to magnum API for now i would go with a CREATE_FAILED and a reason and if the state can be changed via the API we would be good.

okozachenko1203:
> That is a good to know because i am pretty sure we will have users which keep clusters in failed stated forever which means the controller will retry forever which will place some unnessecary load on all openstack APIs. I like the idea of CREATE_BLOCKED but not sure if it is really a good idea to extend it. If we will not add changes to magnum API for now i would go with a CREATE_FAILED and a reason and if the state can be changed via the API we would be good as soon as the user will add quota.

[...]

PR #122, opened 2023-06-07
feat: Expose underlying k8s resource status to coe cluster status reason

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/64

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -58,6 +58,34 @@ class BaseDriver(driver.Driver):
 
         resources.apply_cluster_from_magnum_cluster(context, self.k8s_api, cluster)
 
+    def _get_cluster_status_reason(self, capi_cluster):
+        capi_cluster_status_reason = ""
+        capi_ops_cluster_status_reason = ""
+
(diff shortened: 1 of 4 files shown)


magnum_cluster_api/driver.py
```diff
@@ -100,14 +110,26 @@ def update_cluster_status(self, context, cluster, use_admin_ctx=False):
...
+                cluster.status_reason = (
+                    "Cluster status reason: %s. OpenstackCluster status reason: %s"
+                    % (
+                        resources.get_last_cluster_event(self.k8s_api, cluster),
+                        resources.get_last_openstack_cluster_event(
+                            self.k8s_api, cluster
+                        ),
+                    )
+                )
+                cluster.save()
```
mnaser, line 132: since this code is repeated, does it make sense to refactor this to a function called `_update_cluster_status` or something along those lines?

mnaser (changes requested):
I think there's some ways we can clean this up in the first round of review.

okozachenko1203:
> I think there's some ways we can clean this up in the first round of review.

@mnaser all comments resolved now

Issue #119, opened 2023-06-01
Add labels for controlling OpenID auth


mnaser:
It would be useful to have labels to control the OIDC config for the Kubernetes cluster, such as the following:

- `oidc_client_id`
- `oidc_issuer_url`
- `oidc_username_claim`
- `oidc_username_prefix`
- `oidc_groups_claim`
- `oidc_groups_prefix`

We should also document all of those labels as well for users.

PR #129, opened 2023-06-21
feat: Support OpenID Connect for kube-api auth

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/119

### DIFF
--- a/docs/user/labels.md
+++ b/docs/user/labels.md
@@ -149,37 +149,17 @@ deployment process.
 
 ## OIDC
 
-* `oidc_enabled`
+* `oidc_issuer_url`
 
-   Enable OpenID Connect for the kube-api login.
+   The URL of the OpenID issuer, only HTTPS scheme will be accepted. If set, it
+   will be used to verify the OIDC JSON Web Token (JWT).
 
-   Default value: `false`
+   Default value: ``
 
 * `oidc_client_id`
 
    The client ID for the OpenID Connect client, must be set if `oidc_issuer_url`
    is set.
-
-   Default value: ``
-
-* `oidc_groups_claim`
-
-   If provided, the name of a custom OpenID Connect claim for specifying user
-   groups. The claim value is expected to be a string or array of strings.
-
-   Default value: ``
-
-* `oidc_groups_prefix`
-
-   If provided, all groups will be prefixed with this value to prevent conflicts
-   with other authentication strategies.
-
-   Default value: ``
-
-* `oidc_issuer_url`
-
-   The URL of the OpenID issuer, only HTTPS scheme will be accepted. If set, it
-   will be used to verify the OIDC JSON Web Token (JWT).
 
    Default value: ``
 
@@ -196,6 +176,20 @@ deployment process.
    clashes. To skip any prefixing, use the default value.
 
    Default value: `-`
+
+* `oidc_groups_claim`
+
+   If provided, the name of a custom OpenID Connect claim for specifying user
+   groups. The claim value is expected to be a string or array of strings.
+
+   Default value: ``
+
+* `oidc_groups_prefix`
+
+   If provided, all groups will be prefixed with this value to prevent conflicts
+   with other authentication strategies.
+
+   Default value: ``
 
 ## OpenStack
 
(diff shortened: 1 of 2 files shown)


mnaser (changes requested):
This works great, but I have some ideas around cleaning up the UI:

Since `oidc_issuer_url` is an absolute requirement, no matter what, it will always be set if you're using OpenID connect, we can probably use that as a label to decide if it's on or not (and not rely on `oidc_enabled` instead).

docs/user/labels.md
```diff
@@ -147,6 +147,56 @@ deployment process.
 
    Default value: `true`
 
+## OIDC
+
+* `oidc_enabled`
+
+   Enable OpenID Connect for the kube-api login.
+
+   Default value: `false`
```
mnaser, line 156: This works great, but I have some ideas around cleaning up the UI:

Since `oidc_issuer_url` is an absolute requirement, no matter what, it will always be set if you're using OpenID connect, we can probably use that as a label to decide if it's on or not (and not rely on `oidc_enabled` instead).

docs/user/labels.md
```diff
@@ -147,6 +147,56 @@ deployment process.
...
+   with other authentication strategies.
+
+   Default value: ``
+
+* `oidc_issuer_url`
+
+   The URL of the OpenID issuer, only HTTPS scheme will be accepted. If set, it
+   will be used to verify the OIDC JSON Web Token (JWT).
+
+   Default value: ``
```
mnaser, line 184: Let move this to the first item in the OIDC section.

docs/user/labels.md
```diff
@@ -147,6 +147,56 @@ deployment process.
...
+   Enable OpenID Connect for the kube-api login.
+
+   Default value: `false`
+
+* `oidc_client_id`
+
+   The client ID for the OpenID Connect client, must be set if `oidc_issuer_url`
+   is set.
+
+   Default value: ``
```
mnaser, line 163: and keep this one as the second one.

docs/user/labels.md
```diff
@@ -147,6 +147,56 @@ deployment process.
...
+   The URL of the OpenID issuer, only HTTPS scheme will be accepted. If set, it
+   will be used to verify the OIDC JSON Web Token (JWT).
+
+   Default value: ``
+
+* `oidc_username_claim`
+
+   The OpenID claim to use as the user name.
+
+   Default value: `sub`
```
mnaser, line 190: lets make this third

docs/user/labels.md
```diff
@@ -147,6 +147,56 @@ deployment process.
...
+
+* `oidc_username_prefix`
+
+   If provided, all usernames will be prefixed with this value. If not provided,
+   username claims other than 'email' are prefixed by the issuer URL to avoid
+   clashes. To skip any prefixing, use the default value.
+
+   Default value: `-`
```
mnaser, line 198: fourth

mnaser (changes requested):

magnum_cluster_api/resources.py
```diff
@@ -750,6 +750,43 @@ def get_object(self) -> objects.ClusterClass:
...
+                            "schema": {
+                                "openAPIV3Schema": {
+                                    "type": "object",
+                                    "required": [
+                                        "oidcIssuerUrl",
+                                        "oidcClientId",
+                                        "oidcUsernameClaim",
+                                        "oidcUsernamePrefix",
+                                        "oidcGroupsClaim",
+                                        "oidcGroupsPrefix",
```
mnaser, line 765: these all seem to be redundant that they are prefixed with `oidc` since the top level object is `apiServerOIDC`.

Last small thing, could you rename these to be not prefixed:

```suggestion
                                        "issuerUrl",
                                        "clientId",
                                        "usernameClaim",
                                        "usernamePrefix",
                                        "groupsClaim",
                                        "groupsPrefix",
```

magnum_cluster_api/resources.py
```diff
@@ -750,6 +750,43 @@ def get_object(self) -> objects.ClusterClass:
                                 },
                             },
                         },
+                        {
+                            "name": "apiServerOIDC",
```
mnaser, line 754:
```suggestion
                            "name": "openidConnect",
```

mnaser (changes requested):
one last round. :)

Issue #137, opened 2023-06-29
Creating new node group with autoscaling enabled fails


mnaser:
If you've got a cluster where auto-scaling is enabled and you create one (or multiple node groups at the same time), then I've noticed that you can run into the following error:

```
| status             | CREATE_FAILED                                                                                                                                                                        |
| status_reason      | Unexpected error while running command.                                                                                                                                              |
|                    | Command: helm upgrade --namespace magnum-system kube-7j396 /var/lib/openstack/lib/python3.10/site-packages/magnum_cluster_api/charts/cluster-autoscaler/ --install --wait --values - |
|                    | Exit code: 1                                                                                                                                                                         |
|                    | Stdout: ''                                                                                                                                                                           |
|                    | Stderr: 'Error: UPGRADE FAILED: release: already exists\n'                                                                                                                           |
```

I've also seen other times where it will provide an error along the lines of "release already in progress".  It seems that when creating node groups while the Helm release is already being deployed will result in another attempt at creating a Helm release and fail.

I think we need to somehow detect if the release is in progress (using `helm -n magnum-system ls -a`) or if it already exists and apply it, or maybe the apply function should check if the exit code is not 1, and if the stderr includes "release: already exists" or "another operation (install/upgrade/rollback) is in progress"

okozachenko1203:
@mnaser i think we cannot avoid this kind of error as long as we rely on the helmrelease.
How about to render the helm template files inside the mcapi and apply the plain yaml manifests? In this way, we can apply anytime without worrying about conflict or sth.

mnaser:
@okozachenko1203 the concern with this is clean up.. since we run the auto scaler on the control plane, there will be a service account and many other resources that need to be removed so Helm is better at tracking this. 

Perhaps we can just noop if we get stderr that contains those instead of the exception being raised ?

PR #138, opened 2023-06-29
fix: Do not upgrade helmrelease in pending status

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/137


mnaser:
lol strange how only 3.7 fails.. i will let you look into it @okozachenko1203

PR #146, opened 2023-07-03
fix: Use the default volume type when default_boot_volume_type is not…

### LINKED ISSUE #110: Better handling for default volume
The default volume is not properly selected, we should either rely on Cinder to select the default volume if none is specified or fail early.

okozachenko1203:
… set

fix #110


mnaser:
I ran into an issue in https://github.com/vexxhost/magnum-cluster-api/pull/145 where I borrowed the code and brought it in there.

mnaser:
I merged #145 which contains this.

PR #147, opened 2023-07-03
chore: Drop MachineDeployment annotation workaround

### LINKED ISSUE #142: Drop MachineDeployment annotation workaround
Since we've started rolling out CAPI 1.4, we can probably drop the code that worked around this bug:

https://github.com/kubernetes-sigs/cluster-api/pull/7917

We should probably test locally that the metadata properly propagates though.

okozachenko1203:
fix #142


mnaser:
this is good to go. :)

Issue #164, opened 2023-07-14
Add support for adding `magnum.openstack.org/role` role to nodes.


mnaser:
More information can be found here: https://docs.openstack.org/magnum/latest/user/index.html#roles

We should figure out the best way to attach this, if it's not available out of the box it might be good to know/see what labels that Cluster API adds by default.

okozachenko1203:
There is neither a special label nor annotation.
```
  annotations:
    cluster.x-k8s.io/cluster-name: kube-carvi
    cluster.x-k8s.io/cluster-namespace: magnum-system
    cluster.x-k8s.io/machine: kube-carvi-default-worker-wrjbf-5b8588cb4f-tw765
    cluster.x-k8s.io/owner-kind: MachineSet
    cluster.x-k8s.io/owner-name: kube-carvi-default-worker-wrjbf-5b8588cb4f
    csi.volume.kubernetes.io/nodeid: '{"cinder.csi.openstack.org":"3c648120-e4c0-4fc0-b974-c4658e437b01","manila.csi.openstack.org":"kube-carvi-default-worker-infra-t2f8r-kjl88","nfs.csi.k8s.io":"kube-carvi-default-worker->
    kubeadm.alpha.kubernetes.io/cri-socket: unix:///var/run/containerd/containerd.sock
    node.alpha.kubernetes.io/ttl: "0"
    projectcalico.org/IPv4Address: 10.0.0.182/24
    projectcalico.org/IPv4IPIPTunnelAddr: 10.100.161.128
    volumes.kubernetes.io/controller-managed-attach-detach: "true"

  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: m1.medium
    beta.kubernetes.io/os: linux
    failure-domain.beta.kubernetes.io/region: RegionOne
    failure-domain.beta.kubernetes.io/zone: nova
    kubernetes.io/arch: amd64
    kubernetes.io/hostname: kube-carvi-default-worker-infra-t2f8r-kjl88
    kubernetes.io/os: linux
    node.kubernetes.io/instance-type: m1.medium
    topology.cinder.csi.openstack.org/zone: nova
    topology.kubernetes.io/region: RegionOne
    topology.kubernetes.io/zone: nova

```

okozachenko1203:
```sh
$ o coe nodegroup list 4e1910e1-35d1-4d99-913d-2ee5765156b2
[...]

PR #167, opened 2023-07-18
feat: Add labels for nodegroup name and role name

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/164


okozachenko1203:
image
role field added for master and worker correctly also with our new labels.

PR #174, opened 2023-07-21
chore: Update user doc related to nodegroup roles

okozachenko1203:
Current doc about the nodegroup role is not enough to understand for users who are expecting the old labels like `magnum.openstack.org/role`.
More verbose description added.

### DIFF
--- a/docs/user/getting-started.md
+++ b/docs/user/getting-started.md
@@ -102,5 +102,36 @@ $ openstack coe cluster upgrade <cluster-name> <cluster-template-name>
 
 ### Node group role
 Roles can be used to show the purpose of a node group, and multiple node groups can be given the same role if they share a common purpose.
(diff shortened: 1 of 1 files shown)


okozachenko1203:
mcapi doesn't support `magnum.openstack.org/role` label. Instead of that, it supports k8s-native labels such as `node-role.kubernetes.io/` and `node.cluster.x-k8s.io/`.

The goal of labels is to filter the nodes using the role infor provided during the coe node group creation at the openstack level.
And also to schedule applications(pods) on the specific-role nodes

This goal can be met as long as we propogate openstack nodegroup role infor to k8s level regardless of which label string we use, isn't it?

Also our new labels are much better.
Openstack role infor is now set as K8s node role without a doubt because we use k8s native labels for role setting. So it is much more better than the `magnum.openstack.org/*` labels which is not related to k8s.

PR #184, opened 2023-08-04
Create cluster with existing private network

migmeneses:


migmeneses:
Maybe this over surfaces a lot in the getting started guide

okozachenko1203:
@migmeneses Thanks for your detailed doc.
But IMHO, we don't need to be verbose with contents related to basically magnum itself cause you know it is a specific magnum-driver project. So the things specific to this driver are reasonable in the doc.

I think there is no driver-specific explanation in this topic

Issue #191, opened 2023-08-18
Autoscaling is failing in 1.27 clusters


okozachenko1203:
To support 1.27, we need to bump autoscaler chart to 9.29.1.

okozachenko1203:
Autoscaler1.27 is failing to get the list of `machinepools` because of perm lack. https://github.com/kubernetes/autoscaler/commit/2c7c6c48fbca4330649e7db5bb5a956ce6b8b2d6

okozachenko1203:
@mnaser 
- i think we have to consider the version upgrade for not only autoscaler but also other addons (like csi, ccm, cni, etc) which are maintained by mcapi driver.
- Also i have another concern. It is that the available versions supported by mcapi could be limited by those addons' version compatibility matrix. To mitigate that, we could need to keep multiple versions of addons in mcapi.

What do you think?

fnpanic:
Checking for depracted APIs can be done with pluto,KubePug,KubeNoTrouble and kubent. This will help prevent this problems in the future and can be integrated in the CI workflow.

Checking the deprecation guide is a good idea anyway before declaring a version as stable:

https://kubernetes.io/docs/reference/using-api/deprecation-guide/

mnaser:
@okozachenko1203 i think this is the use case to look at this:

https://github.com/vexxhost/magnum-cluster-api/issues/74

if we implement the addons to be managed, this means we can use helm charts to deploy them and we can just use helm values to manage state/etc.

perhaps we can work on that effort first, and then we start managing those cluster addons at that point using the provider, and potentially even start moving the autoscaler into the cluster itself too and manage it via an addon provider.

PR #192, opened 2023-08-21
fix: Update cluster-autoscaler chart version to support k8s 1.27

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/191


okozachenko1203:
after some investigation, i found that the images built inside the gh workflow is not functioning correctly.

okozachenko1203:
@mnaser I filed an issue for CI failure https://github.com/vexxhost/magnum-cluster-api/issues/194
So we can merge this PR. I tested on my lab and also inside the runner using pre-built images with tmate session.

okozachenko1203:
@mnaser it is failing because of missing images 
https://github.com/vexxhost/magnum-cluster-api/actions/runs/7002925191
Could you upload those artifacts to the swift container when you have a chance?

Issue #180, opened 2023-08-02
Bind address for kube scheduler and controller manager


adamhackl:
In order for Prometheus to properly scrape the health endpoints of kube-scheduler and kube-controller-manager, the --bind-address needs to be set to 0.0.0.0 or the control plane node IP. Clusters are currently deploying with the default of 127.0.0.1, which does not work in this case. Is there any way we can change the default setting or configure an override so clusters deploy this way instead of having to update the configuration post-deploy?

Thank you!

adamhackl:
That would be a great solution for us. Could you also do something similar for the metricsBindAddress for kube-proxy?

Thank you!

mnaser:
I think this needs to wait till #182 is done, without it, none of those ports will be exposed inside the cluster.

Mazorius:
@mnaser that should be also implemented for etcd and [kube-proxy](https://github.com/vexxhost/magnum-cluster-api/issues/180#issuecomment-1663859576) … as they are also configured to 127.0.0.1 and can therefore not be scraped by Prometheus.

- etcd is also an control-plane manifest.
- kube-proxy is a daemon-set were the bind address must be changed.

etcd can be handled with this isssue. If kube-proxy should be tracked with a second can be the case.

#182 seems to be done already.

mnaser:
Alright, so I went over the different components:

- `kube-apiserver` listens on the node IP by default, nothing to do.
- `kube-controller-manager` listens on `127.0.0.1:10257`
- `kube-scheduler` listens on `127.0.0.1:10259`
- `kube-proxy` listens on `127.0.0.1:10249`
- `etcd` listens on `127.0.0.1:2381`

from a security perspective, if we change those to `0.0.0.0` -- it only exposes it within the internal network the cluster lives in, as well as it is protected behind authorization which means someone has to be inside the network that the cluster lives in AND has access credentials that can pull in metrics / talk to metrics, and if you have that kind of credentials, you can probably spin up pods and do many other `$things`.

in this case, to verify, after updating `kube-controller-manager`:

```
root@kube-9b7kp-control-plane-gmjzf-hfkcw:/host/etc/kubernetes/manifests# curl 10.0.0.163:10257
Client sent an HTTP request to an HTTPS server.
root@kube-9b7kp-control-plane-gmjzf-hfkcw:/host/etc/kubernetes/manifests# curl https://10.0.0.163:10257
curl: (60) SSL certificate problem: self-signed certificate in certificate chain
[...]

PR #197, opened 2023-09-01
feat: Update bind address for k8s components

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/180


mnaser:
> Given the security implications (even if low risk), I think it'd be nice to have just a quick reminder in some docs about this being the default. WDYT?

https://github.com/vexxhost/magnum-cluster-api/issues/180#issuecomment-1699764893

I've done that extensively here.

Issue #207, opened 2023-09-14
bug: CA file is not passed to CloudConfigSecret maintained by this driver


okozachenko1203:
Parent issue https://github.com/vexxhost/magnum-cluster-api/issues/204

PR #208, opened 2023-09-14
fix: Add cacert in cloud config for child clusters

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/207


okozachenko1203:
@mnaser this PR adds cacert in the cloud config for coe clusters.
For capo, there are 2 options https://cluster-api-openstack.sigs.k8s.io/clusteropenstack/configuration.html#ca-certificates
I think global cert config meets our demand, but it requires CAPO controller patch. It is out of mcapi driver scope, rather atmosphere scope.

okozachenko1203:
> @mnaser this PR adds cacert in the cloud config for coe clusters. For capo, there are 2 options https://cluster-api-openstack.sigs.k8s.io/clusteropenstack/configuration.html#ca-certificates I think global cert config meets our demand, but it requires CAPO controller patch. It is out of mcapi driver scope, rather atmosphere scope.
> 
> If we choose the per-cluster config, it is said that we have to set `OPENSTACK_CLOUD_CACERT_B64` but i am not sure where we have to set this env var. I raised a question in capo slack channel. Lets see how they respond

ok, we set cacert already for capo

PR #212, opened 2023-09-22
ci: make capi+capo version configurable

mnaser:
Closes https://github.com/vexxhost/magnum-cluster-api/issues/199

PR #214, opened 2023-09-22
fix: Install magnumclient cli in hack script

okozachenko1203:


mnaser:
I added this change into #212

PR #226, opened 2023-09-28
fix: use capi to determine health

### LINKED ISSUE #220: Use CAPI for health checks
At the moment, we're using the built in method provided by the upstream Magnum to monitor cluster health. 

This works in all scenarios when running in something like Atmosphere when all components are in Kubernetes, but in an environment like OSA, it will fail because the conductor cannot reach the proxy endpoint slices. 

The cleaner approach would be to instead to replace the monitor with a custom one that gets the status of the KubeadmControlPlane and all MachineDeployment resources.  This avoids us from dialing out to the cluster and reflects what CAPI really thinks of the cluster.

@jrosser let me know if I missed something from our discussion.

mnaser:
Closes #220

### DIFF
--- a/magnum_cluster_api/monitor.py
+++ b/magnum_cluster_api/monitor.py
@@ -12,27 +12,45 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-from magnum.conductor import k8s_api as k8s
-from magnum.drivers.common import k8s_monitor
+from magnum.conductor import monitors
+from magnum.objects import fields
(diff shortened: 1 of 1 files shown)


jrosser:
Applied this to my OSA AIO and I get CREATE_COMPLETE but heath status in `openstack coe cluster list` says "None"

mnaser:
@jrosser working on this right now, just wanted to get the flatcar support landed. :)

PR #248, opened 2023-11-15
feat: Fetch node count from md object when autoscaling enabled

okozachenko1203:


magnum_cluster_api/driver.py
```diff
@@ -323,7 +323,9 @@ def update_nodegroup_status(self, context, cluster, nodegroup):
                 nodegroup.status = f"{action}_COMPLETE"
             elif phase in ("Failed", "Unknown"):
                 nodegroup.status = f"{action}_FAILED"
-
+            # Note: Fetch node_count from md object when autoscaling enabled
+            if utils.get_auto_scaling_enabled(cluster):
+                nodegroup.node_count = md.obj["spec"]["replicas"]
```
mnaser, line 328: The issue is that this will only be called if the cluster status is `*_IN_PROGRESS` which means it will be out of sync.

I think the better place is to move this to the health status updates

mnaser (changes requested):

magnum_cluster_api/driver.py
okozachenko1203: yeah, i already found where i have to poll.
I never knew we have our own `monitor` in the driver already

okozachenko1203:
this has been tested in my local lab with the following step
- deploy a coe cluster autoscaling enabled with min 2 and max4
- after the cluster is ready, resize the node count of the default worker nodegroup to 6
- the md replicas was reset to 2 by capi driver corrrectly
- the nodegroup replica was retrieved from md replicas by capi driver correctly

PR #253, opened 2023-11-16
chore: Build new k8s version images

okozachenko1203:
/build-new-image


mnaser:
Let's pull out the 1.28 tests for now, we should do that separately.  I think we'd like to drop EOL'd versions.

mnaser:
@okozachenko1203 good work here but can you please update the image loader script to also include images for those new versions so our 'registry' image contains everything? :)

Thanks

PR #267, opened 2023-12-07
feat: Add environment variable to set proxy port

jrosser:
The proxy selects a random port by default but this causes difficulty in an environment where network traffic is controlled by a firewall, either host iptables or a physical appliance.

This patch adds an environment variable, PROXY_PORT, which allows the port to be pre-determined and made consistent.

### DIFF
--- a/magnum_cluster_api/proxy/utils.py
+++ b/magnum_cluster_api/proxy/utils.py
@@ -40,6 +40,7 @@ def get_default_ip_address():
         )
         return ip_address
 
+
 def find_free_port(port_hint=0):
     with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
         try:
--- a/magnum_cluster_api/proxy/utils.py
+++ b/magnum_cluster_api/proxy/utils.py
@@ -40,9 +40,11 @@ def get_default_ip_address():
         )
         return ip_address
 
-
-def find_free_port():
+def find_free_port(port_hint=0):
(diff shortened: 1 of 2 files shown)


okozachenko1203 (changes requested):
I think it is better to validate if the specified port number is idle or not. If that is not idle, then find a free port.
How about to add a new utils function to get haproxy port number.

jrosser:
The usual way to test for port in use is to attempt to bind a socket and check for EADDRINUSE - do we want to be opening/closing that socket in the proxy service process when eventually it is to be used in haproxy?

magnum_cluster_api/proxy/utils.py
```diff
@@ -39,9 +40,12 @@ def get_default_ip_address():
...
-
-def find_free_port():
+def find_free_port(port_hint=0):
     with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
-        s.bind(("", 0))
+        try:
+            s.bind(("", port_hint))
+        except socket.error as e:
+            if e.errno == errno.EADDRINUSE:
+                s.bind(("", 0))
```
okozachenko1203, line 49:
```suggestion
        except OSError:
            s.bind(("", 0))
```
How about to try a free port if bind failed regardless of the errno?. And i think socket.error was deprecated?

jrosser: Done

Issue #9, opened 2022-10-26
Add support for `keystone_auth_enabled`


mnaser:
The main implementation is here: https://github.com/openstack/magnum/blob/2a7ab5e9a37565760fb57fb00f5604f1486cdfb1/magnum/drivers/common/templates/kubernetes/fragments/enable-keystone-auth.sh

We can replace that with a simple ClusterResourceSet.

PR #297, opened 2024-01-22
feat: support keystone-auth

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/9


okozachenko1203:
This has been tested on local lab

mnaser (changes requested):
I think we can take this as an opportunity to start cleaning up all of our manifests which are bundled.  I think it's very difficult to manage them at the moment and I would like to suggest the following:

1. Create a `charts` folder and start bundling the charts in there, if there is no chart, we build it (in this case for this)
2. Since we already have Helm on the system, we can simply use `helm template` with a custom values file instead of using Jinja2
3. With that in place, we're able to simply just bump the chart data in here and easily change the overrides by relying on `values` instead of raw manifest files.
4. The templated value is what we "feed" into the ClusterResourceSet, and it also makes it easier for the long run to transition to the addon providers as well since we will have the charts ahead of time.

Can you make changes to this to accomodate for that?

mnaser:
@okozachenko1203: this assumes that `magnum-conductor` can reach the cluster directly.  In Atmosphere, it can. However, when users are using OSA or Kolla, that is likely not the case, so we'll have to rely on CAPO for the cluster to be reachable (especially in scenarios when the cluster proxy is running).

magnum_cluster_api/resources.py
```diff
@@ -331,6 +331,65 @@ def get_object(self) -> pykube.ConfigMap:
...
+                        ),
+                        values={
+                            "conf": {
+                                "auth_url": auth_url
+                                + ("" if auth_url.endswith("/v3") else "/v3"),
+                                "ca_file": (
+                                    "/etc/kubernetes/cloud_ca.crt"
+                                    if utils.get_cloud_ca_cert()
+                                    else ""
+                                ),
```
mnaser, line 359: refactor into helm chart value

magnum_cluster_api/resources.py
```diff
@@ -331,6 +331,65 @@ def get_object(self) -> pykube.ConfigMap:
...
+                                                "path": "/etc/kubernetes/cloud_ca.crt",
+                                                "type": "File",
+                                            },
+                                            "name": "cloud-ca-cert-volume",
+                                        }
+                                    ]
+                                    if utils.get_cloud_ca_cert()
+                                    else []
+                                ),
+                            },
```
mnaser, line 387: move logic to helm chart

magnum_cluster_api/utils.py
```diff
@@ -395,3 +396,19 @@ def convert_to_rfc1123(input: str) -> str:
     :rtype: str
     """
     return re.sub(r"[^a-zA-Z0-9]+", "-", input).lower()
+
+
+def get_keystone_auth_default_policy(cluster: magnum_objects.Cluster):
```
mnaser, line 401: lets make this return a dicft

mnaser (changes requested):

mnaser (approved):

PR #313, opened 2024-02-27
fix: respect `availability_zone` for control plane

### LINKED ISSUE #257: Enable `availability_zone` label for clusters
https://docs.openstack.org/magnum/latest/user/index.html#high-availability

We need to make sure that all the components such such as CSI, etc.

mnaser:
Fixes #257


mnaser:
@okozachenko1203 the patch in its current state breaks clusters without an az specificed... not sure the best way to do fix this..

PR #320, opened 2024-03-12
Multiple control plane availability zones

0x00ace:
Suggesting an enhanced control plane availability zones support as described in [docs](https://cluster-api-openstack.sigs.k8s.io/api/v1beta1/api#infrastructure.cluster.x-k8s.io/v1beta1.OpenStackCluster)

Using label control_plane_availability_zones="az1,az2,azX" we can select multiple availability zones.


0x00ace:
@mnaser could you please consider this PR?

mnaser:
@0x00ace will do, I'll be merging this right after my work in moving us to Zuul CI which will validate it proper proper (and with RockyLinux)

mnaser:
@0x00ace Zuul CI is in, I just merged this with main to see how it looks like (it'll test again all 4 operating systems :])

magnum_cluster_api/resources.py
```diff
@@ -2556,6 +2564,12 @@ def get_object(self) -> objects.Cluster:
                                     self.cluster, "keystone_auth_enabled", True
                                 ),
                             },
+                            {
+                                "name": "controlPlaneAvailabilityZones",
+                                "value": utils.get_cluster_label(
```
mnaser, line 2569: We've moved away from `get_cluster_label` because Magnum natively copies in all the labels to `cluster.labels`, so you can just `cluster.labels.get("...", "")` at this point.

mnaser (changes requested):

magnum_cluster_api/resources.py
0x00ace: Done

mnaser (approved):

PR #324, opened 2024-03-15
fix: stop unexpected rollouts

mnaser:
When making any node group changes, we were previously
making an entire change of all the cluster resources.  With
this change, we're instead simply making the small changes
in the node group which should avoid unexpected full rollouts
of node groups when changing min/max for autoscaling or
resizing a cluster.

### DIFF
--- a/docs/developer/cluster-topology.md
+++ b/docs/developer/cluster-topology.md
@@ -1,0 +1,12 @@
+# Cluster Topology
+
+The Cluster API driver for Magnum makes use of the Cluster topology feature of the Cluster API project.  This allows it to delegate all of the work around building resources such as the `OpenStackCluster`, `MachineDeployments` and everything else managed entire by the Cluster API instead of the driver creating all of these resources.
+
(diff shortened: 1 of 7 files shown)


okozachenko1203:
As long as the coe cluster upgrade with the same coe clustertemplate is supported, I think there is no downside in the approach that will trigger capi clusterclass update only when coe cluster upgrade requested.
If coe cluster upgrade will require a different coe clustertemplate from the current one, users have to create a twin coe cluster template for clusterclass update. If so, we will have more coe clustertemplates, and will end up with having the same number of coe clsutertemplates as the number of coe clusters in the worst case even though we delete the old coe clustertemplates which no coe cluster refers to.
But fortunately, magnum api allows cluster upgrade with the current coe clustertemplate like heat does with `--existing` flag, and we can go with this approach.

mnaser:
Nice, I didn't know that @okozachenko1203 so this is the best of both worlds. 

I will also try and document this a little bit deeper as well in the documentation as well.

PR #329, opened 2024-03-18
doc: add docs for etcd volume

robincron:
add documentation on the new labels added in https://github.com/vexxhost/magnum-cluster-api/commit/544cb77a4a2101dc2af6ee940ab05968f015fc1a


mnaser:
those are for the ClusterClass so those are the internal values.  I've made a commit here that updated things:

https://vexxhost.github.io/magnum-cluster-api/user/labels/

PR #340, opened 2024-03-22
fix: remove mutable default args in generate_machine_deployments_for_…

okozachenko1203:
…cluster

fix https://github.com/vexxhost/magnum-cluster-api/issues/339

### DIFF
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -2144,13 +2144,17 @@ def mutate_machine_deployment(
     context: context.RequestContext,
     cluster: objects.Cluster,
     node_group: magnum_objects.NodeGroup,
-    machine_deployment: dict = {},
+    machine_deployment: dict = None,
 ):
     """
     This function will either makes updates to machine deployment fields which
     will not cause a rolling update or will return a new machine deployment
     if none is provided.
     """
+    # NOTE(okozachenko1203): Initialize as an empty dict if not provided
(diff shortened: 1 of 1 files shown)


mnaser:
@okozachenko1203 please create functional tests to handle this.

The best idea to test the functional tests locally is by simply creating a kind cluster, then install capi/capo (use the scripts in hack/) and look at Jenkinsfile.

okozachenko1203:
> @okozachenko1203 please create functional tests to handle this.
> 
> The best idea to test the functional tests locally is by simply creating a kind cluster, then install capi/capo (use the scripts in hack/) and look at Jenkinsfile.

I think the only functional test scenario for this issue is the cluster upgrade because `generate_machine_deployments_for_cluster` is only called when a cluster is created and upgraded.

let me add another step in `hack/run-functional-tests.sh` file for this upgrade scenario

okozachenko1203:
so the functional test steps will be as following;
- create a cluster using `kube_tag_old`
- do sonobuoy test
- upgrade the cluster to `kube_tag_new`
- do sonobuoy test

This will take very long time. what do you think if we just run sonobuoy smoke tests? it will speed up

mnaser:
@okozachenko1203 I meant in here actually sorry

https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/tests/functional/test_driver.py

mnaser:
Replaced by #349

PR #341, opened 2024-03-22
chore: add github issue templates

okozachenko1203:


okozachenko1203:
this CI failure is false negative. we can ignore now.

https://github.com/vexxhost/magnum-cluster-api/issues/342

mnaser:
Not valid anymore since we're moved to Zuul.

PR #363, opened 2024-04-19
fix: set certificate expiry day for kcpt

okozachenko1203:


magnum_cluster_api/hacks.py
```diff
@@ -0,0 +1,54 @@
...
+            if "certificatesExpiryDays" not in rollout_before:
+                kcpt.obj["spec"]["template"]["spec"].setdefault("rolloutBefore", {})
+                kcpt.obj["spec"]["template"]["spec"]["rolloutBefore"][
+                    "certificatesExpiryDays"
+                ] = 21
+
+                # NOTE(mnaser): Since the KubeadmControlPlaneTemplate is immutable, we need to
+                #               delete the object and re-create it.
+                kcpt.delete()
+                del kcpt.obj["metadata"]["uid"]
```
mnaser, line 50: @okozachenko1203 I'm concerned about this since we're deleting a resource and creating it again and if we somehow fail to recreate after the delete, the cluster can probably go into a pretty odd state.

Is there a way to avoid the delete by any chance or figure out a way to make this "reliable"?  Also, I think maybe we need to introduce locking to this as well in case multiple delete at the same time, then a create conflict happens at the same time, and then all data is lost.

magnum_cluster_api/hacks.py
```diff
@@ -0,0 +1,54 @@
...
+
+                utils.kube_apply_patch(kcpt)
```
mnaser, line 52: Perhaps we can add retries here?

mnaser (changes requested):

magnum_cluster_api/hacks.py
okozachenko1203: I think we don't need to care about conflicts for both deletion and creation.
`object.delete()` will ignore `ObjectDoesNotExist` error and we do create via server-side apply.
https://codeberg.org/hjacobs/pykube-ng/src/commit/c1ac2c97587249e418d04d91f1297d4ffe3631c3/pykube/objects.py#L167-L180

okozachenko1203: ok, i think we need lock because deletion and creation is done inside one function in a row to avoid the deletion of newly created kcpt by another replica.

okozachenko1203: https://github.com/vexxhost/magnum-cluster-api/pull/363/commits/7d64eb0fb8626d51dc2dbebf8c72cc50ef479213

magnum_cluster_api/hacks.py
okozachenko1203: https://github.com/vexxhost/magnum-cluster-api/pull/363/commits/e5f230fba0e208d28bbb7277ac1c12376a99c738

Issue #4, opened 2022-10-26
Add support for `docker_volume_{size,type}`


mnaser:
At the moment, we cannot use this feature due to the fact that CAPO does not support mounting multiple volumes:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1286

Once that's in place, we can use the following labels to create volumes:

- `etcd_volume_size`
- `etcd_volume_type`
- `docker_volume_size`
- `docker_volume_type`

We can use the following functions to determine the volume type (if size is set):

https://github.com/openstack/magnum/blob/16bdedcf2fe6986c995bd415f4e3c70dac914ada/magnum/common/cinder.py#L25-L37

And we can get the sizes from labels/API objects.

mnaser:
For what it's worth, this isn't that important, since most users will find https://github.com/vexxhost/magnum-cluster-api/issues/5 more important since that gives them BFV instances.

PR #370, opened 2024-05-03
feat: support docker volume type and size

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/4

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -423,11 +423,12 @@ class BaseDriver(driver.Driver):
     def update_nodegroups_status(
         self, context, cluster: magnum_objects.Cluster
     ) -> list[magnum_objects.NodeGroup]:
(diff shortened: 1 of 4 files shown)


okozachenko1203:
need to solve the followings still;
- detect the correct disk name for etcd and docker volume respectively among multiple data volumes, and use preKubeadmCommands https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20200423-etcd-data-disk.md#use-script-to-do-the-etcd-mount-and-append-that-script-to-prekubeadmcommands
- json patches tricks for cases; etcd volume used, docker volume used, both used

okozachenko1203:
recheck

mnaser (changes requested):
I'm not sure why we have multiple cases here instead of simply using json patches that layer up onto each other.

PR #379, opened 2024-05-29
Workaround k8s image-builder not handling recent subiquity installer

jrosser:
Recent subiquity versions leave files in /etc/cloud/ that break networking when the images are booted in openstack.


mnaser:
recheck

mnaser:
recheck

PR #389, opened 2024-06-11
fix tenacity.RetryError on cluster resize

0x00ace:
Without this patch `openstack coe cluster resize` will fail with UPDATE_FAILED

This patch fix tenacity.RetryError: RetryError[<Future at 0xXXXXXXXX state=finished returned MachineDeployment>] when resizing nodegroup

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -22,6 +22,7 @@ from magnum.drivers.common import driver
 from magnum.objects import fields
 from tenacity import (
     Retrying,
+    RetryError,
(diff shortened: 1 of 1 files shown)


mnaser:
@0x00ace I think this change is not necessary but let me know otherwise!

0x00ace:
> hey @0x00ace -- I have a more fundamental fix here!
> 
> #388
> 
> Let me know what you think of it, it improves tracking much more.

@mnaser 
Yep, #388 fix is better, closing this pull request

PR #405, opened 2024-07-01
feat: simplify and generalise get_image()

stuartgrace-bbc:
get_image function in image_utils.py had a hardcoded list of image names and code to modify these for use with a local target registry. It fails an assertion if an unexpected image name is passed in.

This commit simplifies the code and has default functionality for any image not explicitly handled, so that

   old_registry/path/image_name

is converted to

   new_registry/image_name

which will work in almost all cases.

### DIFF
--- a/magnum_cluster_api/image_utils.py
+++ b/magnum_cluster_api/image_utils.py
@@ -56,31 +56,9 @@ def get_image(name: str, repository: str = None):
     if not repository:
         return name
 
-    new_image_name = name
     if name.startswith("docker.io/calico"):
-        new_image_name = name.replace("docker.io/calico/", f"{repository}/calico/")
-    if name.startswith("quay.io/cilium"):
-        new_image_name = name.replace("quay.io/cilium/", f"{repository}/cilium/")
-    if name.startswith("docker.io/k8scloudprovider"):
-        new_image_name = name.replace("docker.io/k8scloudprovider", repository)
-    if name.startswith("registry.k8s.io/sig-storage"):
-        new_image_name = name.replace("registry.k8s.io/sig-storage", repository)
-    if name.startswith("registry.k8s.io/provider-os"):
-        new_image_name = name.replace("registry.k8s.io/provider-os", repository)
-    if name.startswith("gcr.io/k8s-staging-sig-storage"):
-        new_image_name = name.replace("gcr.io/k8s-staging-sig-storage", repository)
-    if new_image_name.startswith(f"{repository}/livenessprobe"):
-        return new_image_name.replace("livenessprobe", "csi-livenessprobe")
-    if new_image_name.startswith("registry.k8s.io/coredns"):
-        return new_image_name.replace("registry.k8s.io/coredns", repository)
-    if new_image_name.startswith("registry.k8s.io/autoscaling"):
-        return new_image_name.replace("registry.k8s.io/autoscaling", repository)
-    if (
-        new_image_name.startswith("registry.k8s.io/etcd")
-        or new_image_name.startswith("registry.k8s.io/kube-")
-        or new_image_name.startswith("registry.k8s.io/pause")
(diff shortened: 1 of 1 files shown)


magnum_cluster_api/image_utils.py
```diff
@@ -56,31 +56,9 @@ def get_image(name: str, repository: str = None):
...
-        return new_image_name.replace("registry.k8s.io/coredns", repository)
-    if new_image_name.startswith("registry.k8s.io/autoscaling"):
-        return new_image_name.replace("registry.k8s.io/autoscaling", repository)
-    if (
-        new_image_name.startswith("registry.k8s.io/etcd")
-        or new_image_name.startswith("registry.k8s.io/kube-")
-        or new_image_name.startswith("registry.k8s.io/pause")
-    ):
-        return new_image_name.replace("registry.k8s.io", repository)
+        return name.replace("docker.io/calico/", f"{repository}/calico-")
```
jrosser, line 60: @stuartgrace-bbc is there a typo here` /calico-` rather than `/calico/` ?

magnum_cluster_api/cmd/image_loader.py
```diff
@@ -57,6 +58,10 @@
     required=True,
     help="Target image repository",
 )
+@click.option(
+    "--manifest",
```
gtirloni, line 62: `--images-manifest` would make it more explicit this is about images and not some other type of manifest (now or in the future). wdyt?

magnum_cluster_api/cmd/image_loader.py
```diff
@@ -79,18 +84,24 @@ def main(repository, parallel, insecure):
...
+        #               registry must be able to support a few different versions
+        #               of Kubernetes since it is possible to have multiple
+        #               clusters running different versions of Kubernetes at the
+        #               same time.
+        images = set(
+            _get_all_kubeadm_images()
+            + _get_calico_images()
+            + _get_cloud_provider_images()
+            + _get_infra_images()
+        )
```
gtirloni, line 104: instead of keeping this code, could we supply a default manifest that's equivalent so people don't have to look into two different places (code and manifest) to figure this out?

jrosser: That would extend the scope of this PR as we would have to keep the default manifest in the source tree, manage its installation to a known location and then load it by default. I am conscious that we are making very little progress toward getting this merged so would like to keep the scope contained.

An extracted default manifest would be missing the images found through running kubeadm so I think that this might need more consideration, and should probably be a follow-up PR if anyone were wanting to do that work.

magnum_cluster_api/image_utils.py
```diff
@@ -56,31 +56,11 @@ def get_image(name: str, repository: str = None):
     if not repository:
         return name
 
-    new_image_name = name
     if name.startswith("docker.io/calico"):
```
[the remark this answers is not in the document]
jrosser: This PR removes all of the hardcoded rules that are not a simple replacement. The remaining ones do not conform to that pattern and as I understand it need to remain for compatibility with the registry/repository structure elsewhere in atmosphere(?). @mnaser may be able to give some more context on why things are like this.

magnum_cluster_api/cmd/image_loader.py
jrosser: @gtirloni we have updated this now to be `--images-manifest`

mnaser:
I apologize this has taken a hot minute to review, I've been meaning to create unit tests to try and make sure we don't regress in those, since to be honest we don't really have a test that uses a `container_infra_prefix`...

jrosser:
@mnaser could you please take a quick look at https://github.com/bbc/magnum-cluster-api/commits/external-manifest/

This should be much lower risk and if it looks basically OK I can create another PR with that different approach.

PR #424, opened 2024-08-22
[MCAPI-1] feat: upgrade capi & capo version

okozachenko1203:
capi 1.6 -> 1.8.4
capo 0.9 -> v0.10.5

To support workload cluster 1.31, need to upgrade capi to 1.8.x To support workload cluster 1.30, need to upgrade capi to 1.7.x


mnaser:
recheck

PR #426, opened 2024-08-29
External manifest for image_loader

jrosser:


okozachenko1203:
@jrosser could you fix the conflict?

PR #433, opened 2024-10-04
keystone_auth: always use public endpoint from tenant cluster

andrewbonney:
I may be misunderstanding, but as far as I can tell the keystone auth component lives entirely in a tenant's cluster, and so must always use the public endpoint to access Keystone.

As we change the capi_client endpoint_type to internal in our deployment, this appears to be contributing to a failure of the keystone auth components in tenant clusters.

If there was a way to get this component to use `/etc/kubernetes/cloud.conf` like other similar components I guess that would be preferable to having this separate definition.


mnaser:
That is correct!

Issue #434, opened 2024-10-07
Keystone auth doesn't work with a custom CA


andrewbonney:
It looks like there is a mismatch between the expectations of the keystoneauth 
Helm chart: https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/charts/k8s-keystone-auth/values.yaml#L20 and the way it is being set up: https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/resources.py#L416 (`ca_file` vs `ca_cert`).

Whilst the Helm chart may work if the file contents are passed in, this seems unnecessary as the nodes should already have this file in `/etc/kubernetes/cloud_ca.crt` and could just mount it as in https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/utils.py#L136.

mnaser:
@okozachenko1203 when you're back, can you please have a look at this and evaluate which fix is the best?

Also, maybe we can try and figure out what kind of CI we can to avoid introducing some kind of regression.  I think long term, it would be better to start avoiding creating files on the host and rely on secrets that are mounted because those we can easily fix/change as opposed to editing files directly on disk that are host mounted.

okozachenko1203:
@andrewbonney thanks for your report.
I'll just fix the key field of helm value. Helm chart renders a secret using ca_cert and mount it into daemonset already.
https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/charts/k8s-keystone-auth/templates/daemonset.yaml#L48-L51
https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/charts/k8s-keystone-auth/templates/daemonset.yaml#L63-L65
https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/charts/k8s-keystone-auth/templates/daemonset.yaml#L77-L80
https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/charts/k8s-keystone-auth/templates/secret-ca.yaml
On the other side, keeping ca crt file on node fs is a mcapi-specific implementation so it would be good to keep k8s-keystone-auth helm chart as standalone for loosely coupling.

andrewbonney:
No problem. I've closed #435 and will leave #436 for review.

PR #435, opened 2024-10-07
keystone_auth: fix providing CA file as path

andrewbonney:
Fixes https://github.com/vexxhost/magnum-cluster-api/issues/434

Not yet tested in an environment which doesn't have a custom CA, but I can't see why this would change the behaviour there.

### DIFF
--- a/magnum_cluster_api/charts/k8s-keystone-auth/templates/daemonset.yaml
+++ b/magnum_cluster_api/charts/k8s-keystone-auth/templates/daemonset.yaml
@@ -45,7 +45,7 @@ spec:
             - {{ .Values.conf.auth_url }}
             - --sync-configmap-name
             - {{ include "k8s-keystone-auth.fullname" . }}-sync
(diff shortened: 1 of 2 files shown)


mnaser:
@andrewbonney Is this meant to be in an environment where the ca_file already exists on the image?

andrewbonney:
Not in the image. The CA file is loaded into tenant cluster hosts via https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/resources.py#L2584 and used successfully by the cloud controller manager where a deployment has its own CA. There's just a mismatch in the variable naming for keystone_auth between the Helm chart and the resources.py file: https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/resources.py#L416

andrewbonney:
I note however that get_openstack_ca() in Magnum returns an empty string rather than None if it's not set, so unless Helm's 'empty' works for that too this may need a slight modification where the CA cert isn't added explicitly.

PR #441, opened 2024-10-17
feat: Add insecure_registry_url label option & Update containerd sett…

harunalbayrak:
Adding insecure_registry_url label and Updating containerd settings to download the images with an insecure registry.

### DIFF
--- a/magnum_cluster_api/utils.py
+++ b/magnum_cluster_api/utils.py
@@ -198,13 +198,18 @@ def get_cluster_floating_ip_disabled(cluster: magnum_objects.Cluster) -> bool:
     return not get_cluster_label_as_bool(cluster, "master_lb_floating_ip_enabled", True)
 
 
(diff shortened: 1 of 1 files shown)


mnaser:
Hey @harunalbayrak !  Thanks for your contribution.

The Magnum API seems to ship `insecure_registry` as an attribute in the cluster template, what do you think about possibly using that instead?

harunalbayrak:
> Hey @harunalbayrak ! Thanks for your contribution.
> 
> The Magnum API seems to ship `insecure_registry` as an attribute in the cluster template, what do you think about possibly using that instead?

You are welcome @mnaser. 

The Magnum API ships insecure_registry as an attribute but it works only with docker. So, Magnum API doesn't handle config.toml file which is used in containerd.  Since containerd is used in Cluster API, the insecure_registry attribute doesn't work. 

I thought that it was the best solution for this situation. Do you have any advice?

PR #455, opened 2024-11-12
Add manifests for relevant calico versions

noonedeadpunk:
Starting with k8s 1.28 expected calico version is 3.28. While we should
dynamicially retrieve manifests for Calico, it needs slighly bigger
effort to implement if consider environments without Internet access.
Alternatively, image version can be templated in manifests, but there're
more changes to them happening from time to time, so still better to
use separate ones.

So as a workaround adding fresh calico manifests to the repo directly.

### DIFF
--- a/docs/user/labels.md
+++ b/docs/user/labels.md
@@ -144,8 +144,11 @@ is often accomplished by deploying a driver on each node.
 
    The version of the Calico container image to use when bootstrapping the
    cluster.
+   Please note, that in case of selecting version out of the supported range,
+   you will need to supply a manifest for it.
 
    Default value: `v3.24.2`
+   Supported values: `v3.24.2`, `v3.25.2`, `v3.26.5`, `v3.27.4`, `v3.28.2`, `v3.29.0`
 
 ## Container Storage Interface (CSI)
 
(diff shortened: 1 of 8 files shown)


mnaser:
@okozachenko1203 if this looks good to you, maybe we merge this?

mnaser:
@okozachenko1203 I'm also not opposed to bumping the default tag too

noonedeadpunk:
> @okozachenko1203 I'm also not opposed to bumping the default tag too

What I've learned, is that calico version should match appropriate k8s version.
So I'd suggest bumping it together with default `kube_tag`.
So makes sense to cover it with a follow up PR, imo.

noonedeadpunk:
PS: I was talking manifests simply by wget of https://github.com/projectcalico/calico/blob/master/manifests/calico.yaml

That's why I think ideally it should try to dynamically fetch in the future, but have not figured out yet on how to handle that in isolated environments

noonedeadpunk:
Actually wonder if this one could be an alternative solution for what I'm trying to do: https://github.com/vexxhost/magnum-cluster-api/pull/426/files

As now when I'm looking at failing test and https://github.com/vexxhost/magnum-cluster-api/blob/main/magnum_cluster_api/cmd/image_loader.py#L178-L184 and not sure anymore about how patch this nicely either, as this all seems very hardcoded to me...

okozachenko1203 (approved):

PR #456, opened 2024-11-15
Add a label to control load-balancer monitor creation

noonedeadpunk:
There are usecases when creation of load balancer monitors are required.
Let's add a corresponsive label to control behaviour of monitors
creation.
Default value is set in accordance with CCM.

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/5ed0a04dc732d0e9bf70ddd3beaccdabd3841957/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md?plain=1#L231-L232

### DIFF
--- a/docs/user/labels.md
+++ b/docs/user/labels.md
@@ -129,6 +129,13 @@ is often accomplished by deploying a driver on each node.
    Default value (`amphora` provider): `ROUND_ROBIN`
    Default value (`ovn` provider): `SOURCE_IP_PORT`
 
+* `octavia_lb_healthcheck`
+
+   The Octavia Load Balancer members can be monitored with health monitor.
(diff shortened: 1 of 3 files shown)


mnaser:
@noonedeadpunk we try to stick to upstream labels, and i think the following is what we need to call this one

https://docs.openstack.org/magnum/latest/user/#octavia-lb-healthcheck

mnaser (changes requested):
@noonedeadpunk we try to stick to upstream labels, and i think the following is what we need to call this one

https://docs.openstack.org/magnum/latest/user/#octavia-lb-healthcheck

noonedeadpunk:
Ah, fair, I didn't double-check with magnum docs if this is already existing.

PR #464, opened 2024-11-29
Spawn cinder volumes in the same Availability Zone

noonedeadpunk:
At the moment volumes for workers are gonna be spawn in a random
AZ completely disregarding user request for Availability Zone.

In a design where cross_az_attach is not posisble an attempt
to add a volume to worker has high failure percentage due to random
selection of AZ.

This patch suggests using provided to cluster availability_zone by
supplying `availability` parameter to cinder csi [1]

This ensures that volume will be created in a same zone as workers are
preventing failures where cross_az_attach is False.

[1] https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#supported-parameters


noonedeadpunk:
Ok, now I see how it's wrong... As eventually you might have only 1 AZ in cinder while have multiple in Nova, which will break the behavior who has cross-az attach :(

noonedeadpunk:
I'd say there should be a follow-up patch to add smth like `docker_volume_availability_zone` label to allow users to override default AZ for cinder.

noonedeadpunk:
recheck

noonedeadpunk:
@mnaser @okozachenko1203 it would be really nice to get some input on this one

magnum_cluster_api/resources.py
```diff
@@ -296,6 +296,15 @@ def get_object(self) -> pykube.ConfigMap:
...
             default_volume_type = osc.cinder().volume_types.default()
+            # Default is set in accordance to the CSI:
+            # https://github.com/kubernetes/cloud-provider-openstack/blob/d228854cf58e7b4ed93d5e7ba68ab639450e3449/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#supported-parameters
+            # If allow_availability_zone_fallback is set to False in Cinder
+            # and "nova" AZ is not present in Cinder, CSI request will fail.
+            volume_availability_zone = "nova"
+            if not CONF.cinder.cross_az_attach:
+                volume_availability_zone = self.cluster.labels.get(
+                    "availability_zone", volume_availability_zone
+                )
```
mnaser, line 307: I feel like this is making an assumption that the default AZ is `nova`.

I think what we should instead do is if `[cinder]/cross_az_attach` is set to `false`, we should set the availablity zone -- otherwise, we leave it empty and maintain the existing behaviour.

mnaser (changes requested):

magnum_cluster_api/resources.py
noonedeadpunk: Ok, sure will do. 

Just thought that matching the default value in csi should provide same behavior and simplify a resulting dict update which should be done conditionally then (as I don't think I can set the value to `None` here - it's gonna change the behavior indeed).

noonedeadpunk:
So Zuul jobs are failing with a completely unrelated:
```
+ sh -c curl -fsSL "[https://download.docker.com/linux/ubuntu/gpg"](https://download.docker.com/linux/ubuntu/gpg%22) -o /etc/apt/keyrings/docker.asc
curl: (56) OpenSSL SSL_read: error:0A000126:SSL routines::unexpected eof while reading, errno 0
```

mnaser:
Looks like Docker had issues yesterday, but this is a problem:

```
diff --git a/magnum_cluster_api/resources.py b/magnum_cluster_api/resources.py
index df5f488..de5d78f 100644
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -325,8 +325,9 @@ class ClusterResourcesSecret(ClusterBase):
                                             "availability_zone"
                                         )
                                     }
-                                    if not CONF.cinder.cross_az_attach else {}
-                                )
+                                    if not CONF.cinder.cross_az_attach
+                                    else {}
+                                ),
                             },
                             "reclaimPolicy": "Delete",
                             "volumeBindingMode": "Immediate",
```

noonedeadpunk:
Ok, yeah, makes sense. I just was unable to run `tox -e linters` even after the rebase, as it was failing due to absent binaries which are not included into the list of allowed for installation software to the workstations so far.

noonedeadpunk:
> Btw, I am curious how it will work when we have multiple node groups with different availability_zone labels.

I'm not sure I tested that specifically. My usecase was more or less having 3 independent AZs with independet storages. So basically volume AZ there was matching with AZ for VMs.

The usecase of Cinder having a different AZ from Nova is probably not covered with this one...

PR #470, opened 2025-01-08
Improve status_reason when cluster creation is stalled

jrosser:
If during cluster creation or update, one or more nodes cannot be created by Nova (because, for example, the project runs out of quota, or Nova runs out of suitable hypervisors), the cluster stalls with CREATE_IN_PROGRESS status. The status_reason in this case is not helpful as it only reports the previous step which succeeded.

This patch adds a check of each individual machine status looking for machines which are not ready and reports any reason found. This can result in useful status_reason messages such as
 'error creating Openstack instance ... Quota exceeded for instances'


okozachenko1203:
@jrosser could you rebase your branch? we improved tests in CI so good to see the CI success result before peer-review.

magnum_cluster_api/driver.py
```diff
@@ -220,6 +220,25 @@ def update_cluster_status(
...
+                for cond in machine.obj["status"]["conditions"]:
+                    if (
+                        cond.get("type") == "InfrastructureReady"
+                        and cond.get("status") == "False"
+                    ):
+                        messagetext = cond.get("message")
+                        if messagetext:
+                            cluster.status_reason = messagetext
+                            cluster.save()
+
```
okozachenko1203, line 241: This will overwrite `status_reason` generated by `_get_cluster_status_reason` in the former step.
Current `_get_cluster_status_reason` generates status reason message from `Cluster` and `OpenstackCluster` CRs. 
I recommend appending your code snippet into `_get_cluster_status_reason` function.
You'll need to loop all machines and concatenate messages with machine name. Then append this all-machine message in the final reason message alongside existing ones.

https://github.com/vexxhost/magnum-cluster-api/blob/b2a612f0b7aa6d71d2d390217d443f0f84695f25/magnum_cluster_api/driver.py#L124-L148

okozachenko1203 (changes requested):
@jrosser thanks for your quick response.
I have one comment now.

magnum_cluster_api/driver.py
jrosser: Unfortunately the original author of this patch has now taken retirement and no longer works on my team. Realistically it is going to need someone else to pick this up as I don't have any engineering effort available to work on this for the foreseeable future.

PR #471, opened 2025-01-09
[MCAPI-54] feat: add validation logic for zero boot vol case

okozachenko1203:
when both boot volume size and flavor's root_gb are zero, cluster node vm creation will be failed.
This commit prevents cluster creation with such params.


okozachenko1203:
recheck

okozachenko1203:
recheck

PR #472, opened 2025-01-09
Use calico images from quay.io for workload clusters

jrosser:
Instead of getting these from docker.io and running into rate limits use images from quay.io


mnaser:
Since I cannot edit your PR and wanted to add some follow up unit test, I opened https://github.com/vexxhost/magnum-cluster-api/pull/474

PR #474, opened 2025-01-10
Use calico images from quay.io for workload clusters

mnaser:
- **Use calico images from quay.io for workload clusters**
- **Add unit tests + fix Calico regression**


mnaser:
This is covered in the Rust refactor which comes in https://github.com/vexxhost/magnum-cluster-api/pull/494 -- already there.  I'm hoping to land it before end of this month before the rate limits come in.

PR #476, opened 2025-01-13
fix server group not being deleted after cluster deletion

yaguangtang:


mnaser (changes requested):
I think the issue here is that it's possible this is called either with admin context or with non admin context. Can you set all projects if it's being called with admin context?

Alternatively, is there a way to avoid this whole call when deleting?

magnum_cluster_api/utils.py
```diff
@@ -534,7 +534,7 @@ def get_server_group_id(
 
     # Check if the server group exists already
     osc = clients.get_openstack_api(ctx)
-    server_groups = osc.nova().server_groups.list()
+    server_groups = osc.nova().server_groups.list(all_projects=True)
```
okozachenko1203, line 537:
```suggestion
    server_groups = osc.nova().server_groups.list(all_projects=ctx.is_admin)
```

PR #484, opened 2025-01-28
[MCAPI-57] feat: ensure the cluster upgrade follows the minor/patch upgrade rule

okozachenko1203:


mnaser (changes requested):
Wouldn't this cluster end up in a UPDATE_FAILED state? :(

okozachenko1203:
> Wouldn't this cluster end up in a UPDATE_FAILED state? :(

yes, it'll end up in a failed state.

There are 2 ways to reject update operation keeping the cluster status unchanged.
- add version check in magnum project instead of mcapi
- raise `exception.NotSupported` inside mcapi. it will be detected in magnum conductor but the cluster status will be unchanged

https://opendev.org/openstack/magnum/src/commit/1c3d7d070b60a36ccfb7c753b26f12609d818cec/magnum/conductor/handlers/cluster_conductor.py#L337-L354

okozachenko1203:
I went with the latter option.

PR #494, opened 2025-02-21
feat: Create strong typed resources

mnaser:


magnum_cluster_api/driver.py
```diff
@@ -413,7 +427,6 @@ def delete_cluster(self, context, cluster: magnum_objects.Cluster):
         #               https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/990
         utils.delete_loadbalancers(context, cluster)
 
-        resources.ClusterResourceSet(self.kube_client, cluster).delete()
```
mnaser, line 416: note to self: last thing to land this is to fix the delete here.

PR #501, opened 2025-02-21
fix dns server list issue when specified with more than one

yaguangtang:
In magnum template when specify more than one dns servers,the cluster using that template fail to create due to dns list being regard as string
https://github.com/gophercloud/gophercloud/blob/main/openstack/networking/v2/subnets/requests.go#L125C1-L125C60

### DIFF
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -2048,7 +2048,10 @@ class ClusterClass(Base):
                                             "template": textwrap.dedent(
                                                 """\
                                                     - cidr: {{ .nodeCidr }}
-                                                      dnsNameservers: {{ .dnsNameservers }}
+                                                      dnsNameservers:
+                                                     {{- range .dnsNameservers }}
+                                                        - {{ . }}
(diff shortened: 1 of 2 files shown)


magnum_cluster_api/resources.py
```diff
@@ -2048,7 +2048,10 @@ def get_object(self) -> dict:
                                             "template": textwrap.dedent(
                                                 """\
                                                     - cidr: {{ .nodeCidr }}
-                                                      dnsNameservers: {{ .dnsNameservers }}
+                                                      dnsNameservers:
+                                                      {{- range .dnsNameservers }}
+                                                        - {{ . }}
+                                                      {{- end }}
                                                     """
```
mnaser, line 2055: I think we should create a functional test for this to test the behaviour when 1 or many DNS servers are being included and make sure that it propagates properly.  There are examples of this already

mnaser:
This is covered in the Rust refactor in https://github.com/vexxhost/magnum-cluster-api/pull/501

PR #510, opened 2025-02-23
Add AZ, Flavor, and Floating IP labels for apiServerLoadBalancer

nguyenhuukhoi:
Users can choose AZ, Flavor and specify a Floating IP for their apiServerLoadBalancer. This is useful for multi-AZ systems.

### DIFF
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -1068,12 +1068,27 @@ class ClusterClass(Base):
                                     "provider": {
                                         "type": "string",
                                     },
+                                    "availabilityZone": {
+                                        "type": "string",
+                                    },
+                                    "flavor": {
+                                        "type": "string",
+                                    },                                    
                                 },
                             },
                         },
                     },
                     {
                         "name": "apiServerTLSCipherSuites",
+                        "required": True,
+                        "schema": {
+                            "openAPIV3Schema": {
+                                "type": "string",
+                            },
+                        },
+                    },
+                    {
+                        "name": "apiServerFloatingIP",
                         "required": True,
                         "schema": {
                             "openAPIV3Schema": {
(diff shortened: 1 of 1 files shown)


MaximMonin:
I am also interested in confirming this commit, since I also have a need to bind a kubernetes cluster to an existing Floating IP address. Once the cluster is created there is no way to change cluster.api_address.

mnaser:
@nguyenhuukhoi @MaximMonin i've rebuilt this into a rust feature as we're refactoring a big majority of the project to make it a lot more testable.  i was also able to add unit tests to validate if it works and doesn't work.

if this looks good and passes ci, i'll merge it.

nguyenhuukhoi:
add api server floating ip feature, this failed because regular users dont have permission to create floating ips. We need change 
neutron policy for it "create_floatingip:floating_ip_address: "rule:regular_user"".

MaximMonin:
I have another suggestion regarding this commit. Currently, two lb providers are used: amphora and ovn (although the user can potentially use others). We plan to always use amphora for the master LB, as it is more stable and also has the ability to set a firewall (which can be configured after the cluster creation process). Whereas for kubernetes services, we can use ovn.

There was an idea to always use amphora if flavor is set (it cannot be set for ovn). Or add a new label that overrides it (master_lb_octavia_provider).

mnaser:
> add api server floating ip feature, this failed because regular users dont have permission to create floating ips. We need change 
> 
> neutron policy for it "create_floatingip:floating_ip_address: "rule:regular_user"". 

In that case, I don't think we can support this feature.

What if you provide it with a floating ip that already exists, wouldn't that workaround the issue ?

mnaser:
> I have another suggestion regarding this commit. Currently, two lb providers are used: amphora and ovn (although the user can potentially use others). We plan to always use amphora for the master LB, as it is more stable and also has the ability to set a firewall (which can be configured after the cluster creation process). Whereas for kubernetes services, we can use ovn.
> 
> 
> 
> There was an idea to always use amphora if flavor is set (it cannot be set for ovn). Or add a new label that overrides it (master_lb_octavia_provider).

Indeed, you're right about this, we "mirrored" the behaviour that Magnum did with the original Heat behaviour but we can diverge here I think. 

Do you mind opening a seperate issue for this though?

nguyenhuukhoi:
It just cannot create cluster and It doest impact our system. and if your
system doest allow this permission. I think it would be ok. I added this
feature for private clouds.
Nguyen Huu Khoi

MaximMonin:
> What if you provide it with a floating ip that already exists, wouldn't that workaround the issue ?

Accoding cluster api - cluster api will use existing ip address, if it is already exists.

nguyenhuukhoi:
Hello @mnaser, could we discuss about this feature. I think it will be very helpful if we have multi AZ.

nguyenhuukhoi:
Hello @okozachenko1203, Could you help me to submit this patch? I will doc labels later. SR for inconvenient. I am not working much with git.

okozachenko1203:
> Hello @okozachenko1203, Could you help me to submit this patch? I will doc labels later. SR for inconvenient. I am not working much with git.

can you give me write permission to your repo then?

okozachenko1203:
> > > Hello @okozachenko1203, Could you help me to submit this patch? I will doc labels later. SR for inconvenient. I am not working much with git.
> > 
> > 
> > can you give me write permission to your repo then?
> 
> Hello, did u access my repo?

hi @nguyenhuukhoi the invitation has been expired.

nguyenhuukhoi:
Hi, I resent it for you.

okozachenko1203:
@nguyenhuukhoi i created a fresh one https://github.com/vexxhost/magnum-cluster-api/pull/798/files
and close this

PR #517, opened 2025-02-27
fix: Cluster upgrade uses the right calls

mnaser:
- **Switch tu updated client**
- **Add update_cluster API call**


mnaser:
closing this until branch is ready so the ci doesnt blow up

PR #523, opened 2025-03-05
force scale-up cluster nodegroup when node_count < min_node_count

MaximMonin:


MaximMonin:
```
chart-vendor --charts-root magnum_cluster_api/charts --config-file .charts.yml --check
non-zero return code
2025/03/05 12:33:57 INFO applying patch chart=cluster-autoscaler version=9.29.1 repository=https://kubernetes.github.io/autoscaler patch=magnum_cluster_api/charts/patches/cluster-autoscaler/001-add-omt-to-clusterrole.patch patch=magnum_cluster_api/charts/patches/cluster-autoscaler/002-enforce-scaleup.patch
2025/03/05 12:33:57 Changed file: magnum_cluster_api/charts/cluster-autoscaler/values.yaml
2025/03/05 12:33:57 uncommitted changes or untracked files found
```
 I cannot understand how to pass --check options. I checked chart-vendor source code which checks git uncommitted files.

mnaser:
@MaximMonin you need to add a patch into the `patches/` folder that would apply into the chart, you can look at the other autoscaler patch to see how it is formed.

MaximMonin:
This commit already uses patches. And it does not work anyway.

Issue #483, opened 2025-01-25
k8s-keystone-auth-policy is missing $PROJECT_ID


mariusleu:
The `k8s-keystone-auth-policy` ConfigMap gets rendered without the project id, thus Keystone tokens can't be used to access the cluster.

In my case it looks like this
```
      "users":{
         "projects":[
            "$PROJECT_ID"
         ],
         "roles":[
            "k8s_admin"
         ]
      }
```

After replacing $PROJECT_ID with the actual project id, then restarting `k8s-keystone-auth` pods, the authentication works well using Keystone.

mnaser:
@mariusleu Can you please push a PR since you've already locked down the exact solution?

PR #525, opened 2025-03-06
fix: replace $PROJECT_ID in the template

mnaser:
Closes: https://github.com/vexxhost/magnum-cluster-api/issues/483


mnaser:
recheck

PR #529, opened 2025-03-11
feat: support virtio-scsi model

mnaser:
- **Remvoe extra patch print**
- **Update feature to support hardwareSCSIModel**
- **Add python variables side**
- **Fix tests**

### DIFF
--- a/magnum_cluster_api/utils.py
+++ b/magnum_cluster_api/utils.py
@@ -460,14 +460,14 @@ def get_image_uuid(image_ref: str, ctx: context.RequestContext):
     return image_obj.get("id")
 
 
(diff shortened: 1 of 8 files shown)


magnum_cluster_api/utils.py
```diff
@@ -460,6 +460,16 @@ def get_image_uuid(image_ref: str, ctx: context.RequestContext):
     return image_obj.get("id")
 
 
+def get_hw_scsi_model(ctx: context.RequestContext, image_ref: str):
+    """Get image uuid from image ref
```
okozachenko1203, line 464:
```suggestion
    """Get image hw_scsi_model info from image ref
```
```suggestion
    """Get image uuid from image ref
```

PR #581, opened 2025-03-20
implementing control plane resize

MaximMonin:


MaximMonin:
recheck

satishdotpatel:
+1 for this patch

Issue #580, opened 2025-03-20
Cluster Upgrade stuck UPDATE_IN_PROGRESS due to Missing cloud.conf in CSI & OCCM


pahrialms:
I recently upgraded CAPI from `v1.8.1` to `v1.9.6` and magnum-cluster-api from `v0.22.0` to `v0.29.0`. After successfully provisioning a new cluster with the latest version, I attempted to upgrade the cluster workload from `clusterclass v0.22.0`. However, the status remains stuck at `UPDATE_IN_PROGRESS`.

Upon checking the `csi-cinder` and `openstack-cloud-controller-manager` pods, I found the following events:

```
FailedMount alert 3m18s (x13 over 13m) kubelet MountVolume.SetUp failed for volume "secret-cinderplugin": hostPath type check failed: /etc/kubernetes/cloud.conf is not a file

Warning FailedMount 76s (x14 over 13m) kubelet MountVolume.SetUp failed for volume "cloud-ca-cert-volume": hostPath type check failed: /etc/kubernetes/cloud_ca.crt is not a file
```

It seems that in the latest version, the `cloud-ca-cert-volume` no longer exists, and `secret-cinderplugin` now uses a Kubernetes Secret instead of `hostPath`.

### **Steps to Reproduce:**
1. Upgrade CAPI from `v1.8.1` to `v1.9.6`.
2. Upgrade magnum-cluster-api from `v0.22.0` to `v0.29.0`.
4. Attempt to upgrade a workload cluster running `clusterclass v0.22.0`.
5. Observe that the upgrade process gets stuck at `UPDATE_IN_PROGRESS`.
6. Check the logs for `csi-cinder` and `openstack-cloud-controller-manager` pods.

### **Expected Behavior:**
- The upgrade process should complete successfully without manual intervention.
- The necessary configurations should be migrated automatically.

### **Actual Behavior:**
- The upgrade process gets stuck.
- The `csi-cinder` and `openstack-cloud-controller-manager` pods fail to mount the required volumes due to missing `cloud.conf` and `cloud_ca.crt`.

### **Workaround:**
- Get the cloud-config secret yaml of the new successfully provisioned cluster, and then deploy it on the cluster in the process of upgrading.
- Manually patched the following components after performing a configuration comparison with the newly provisioned cluster:
  - `csi-cinder-controllerplugin` deployment
  - `csi-cinder-nodeplugin` daemonset
  - `openstack-cloud-controller-manager` daemonset

[...]

PR #587, opened 2025-03-21
fix: use Reconcile strategy for CRS and update CRS during cluster upg…

okozachenko1203:
…rade

fix https://github.com/vexxhost/magnum-cluster-api/issues/580


mnaser:
@okozachenko1203 instead of reconciling everything, I think it's better to just reconcile the cloud controller manager:

https://github.com/vexxhost/magnum-cluster-api/pull/604

PR #638, opened 2025-04-08
use quay.io image for calico

okozachenko1203:


mnaser (changes requested):
Can you update the renovate configuration to do a sed as well.

renovate.json
```diff
@@ -18,7 +18,8 @@
       "postUpgradeTasks": {
         "fileFilters": ["magnum_cluster_api/manifests/calico/v*.yaml"],
         "commands": [
-          "curl -o magnum_cluster_api/manifests/calico/v{{{newValue}}}.yaml https://raw.githubusercontent.com/projectcalico/calico/v{{{newValue}}}/manifests/calico.yaml"
+          "curl -o magnum_cluster_api/manifests/calico/v{{{newValue}}}.yaml https://raw.githubusercontent.com/projectcalico/calico/v{{{newValue}}}/manifests/calico.yaml",
+          "sed -i 's|docker.io|quay.io|g' magnum_cluster_api/manifests/calico/v{{{newValue}}}.yaml"
```
mnaser, line 22: I think Renovate won't like this once we switch to the hosted application because the `postUpgradeTasks` does not support running `curl` commands.

Can you try to think of another way to try and update this?

mnaser (changes requested):

renovate.json
okozachenko1203: @mnaser post-upgrade tasks can only be used on self-hosted Renovate instances. Once we switch to hosted version, then we cannot use not only curl but also this whole post-upgrade.

okozachenko1203: so i think it is impossible to patch manifests via renovate.

mnaser:
I came up with a fancy little workaround :)

https://github.com/vexxhost/multipathd_exporter/blob/main/.github/workflows/ci.yaml#L41-L56

So as long as the pre-commit makes sure that this stuff is in place, we should be good, I'll explore how to do that.

mnaser:
Did some magic and we got updates + quay.io conversion :)

https://github.com/vexxhost/magnum-cluster-api/pull/662

Issue #636, opened 2025-04-07
v0.29.3 caused exception during reset or similar: RuntimeError: do not call blocking functions from the mainloop


nguyenhuukhoi:
Hello.

I got this problem when I run magnum about 8h with creating and deleting. clusters normally, it causes magnum cannot update cluster status and we need restart magnum conductor but it will happen again and again:

I test with 0.29.3 then it is ok.

ERROR sqlalchemy.pool.impl.QueuePool [-] Exception during reset or similar: RuntimeError: do not call blocking functions from the mainloop

This is full log:

https://paste.sh/Ff3F5eMT#BUlO3ncE3MGEy3xHJcDkdWua

Openstack:2024.02
Ubuntu: 22.04
Deployment Tool: Kolla Ansible
Driver: Magnum-Cluster-API(0.29.3)

Thank you much.

nguyenhuukhoi:
v0.29.2 still has this problem but it happen sometimes not much as 0.29.3

nguyenhuukhoi:
with 0.29.2. magnum conductor stucked after
`request /var/lib/kolla/venv/lib/python3.12/site-packages/kubernetes/client/rest.py:235
2025-04-07 23:27:45.655 7 DEBUG kubernetes.client.rest [-] response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"leases.coordination.k8s.io \"cluster-abe11928-7186-4bec-9075-3f3f9a5491f1\" not found","reason":"NotFound","details":{"name":"cluster-abe11928-7186-4bec-9075-3f3f9a5491f1","group":"coordination.k8s.io","kind":"leases"},"code":404}
 request /var/lib/kolla/venv/lib/python3.12/site-packages/kubernetes/client/rest.py:235
2025-04-07 23:27:45.663 7 DEBUG tower.buffer.worker [-] service.ready=true message=processing request None /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tower-0.5.2/src/buffer/worker.rs:170
[...]

PR #645, opened 2025-04-15
fix: use tpool.Proxy for Rust calls

mnaser:
Closes: https://github.com/vexxhost/magnum-cluster-api/issues/636


mnaser:
@nguyenhuukhoi do you mind checking this one out when you're free? :)

nguyenhuukhoi:
Hello. I dont know why but when I login to magnum_conductor container then remove driver and install it again from source then old clusters replace some old nodes by new ones.

mnaser:
@nguyenhuukhoi Is it possible you have any upgrades in progess or anything like that?  Did you make any upgrades such an upgrade or anything like that?

nguyenhuukhoi:
No, i have only one cluster which i created yesterday. I will dig in it.

mnaser:
ok, let me know if those error messages continue as well or not, i'm hoping the thread pool helped things.

nguyenhuukhoi:
Hello. It is ok at this time. I try to create issue but it is ok. What happen when evenlet will be removed in future?

mnaser:
When eventlet is removed, it will be even better because then we can actually use native async and await constructs and move our code to actually be async in Rust too and avoid these small runtime hacks.

PR #648, opened 2025-04-23
fix: delete bad resources pre-upgrade

mnaser:
- **chore: switch to depythonize**
- **chore: initial upgrade detection**
- **revert pythonize**
- **Add status handling**
- **upgrade handling**
- **initial attempt at client**
- **Add notes**


src/magnum.rs
```diff
@@ -88,6 +148,46 @@ impl Cluster {
...
+        let config =
+            Config::from_custom_kubeconfig(kubeconfig, &KubeConfigOptions::default()).await?;
+        let client = Client::try_from(config).map_err(|e| ClusterError::Kubernetes(e))?;
+
+        // TODO: If the Cluster API driver is running outside of the management cluster and this is an
+        //       isolated cluster, we need to create a port-forward to the API server through the
+        //       management cluster.
+
+        Ok(client)
+    }
```
mnaser, line 189: @okozachenko1203 this is where we can generate a new client here specifically for this cluster, now this works fine out of the box for Atmosphere for all clusters.  It also works for any other deployment method _EXCEPT_ if the cluster is isolated (aka no public floating IP).  This is because the endpoint in the Kubeconfig will be `kube-abcde.magnum-system` which will _not_ resolve here.

The trick will be to detect if this is an isolated cluster, if it is and `kube-abcde.magnum-system` is reachable, then yay!  if it is not, then that means we're running outside the cluster (unfortunate) and we have to spin up a port-forward that we talk to.

this is all really to handle: alternative deployment methods (non-Atmosphere) with isolated clusters.  I think this is gonig to be lower priority to fix to be honest.

src/driver.rs
```diff
@@ -53,9 +54,24 @@ impl Driver {
...
+                //               ones which was created by manifests. We need to clean up the
+                //               legacy resources otherwise it will generate a conflict during
+                //               the upgrade.
+                //
+                //               https://github.com/vexxhost/magnum-cluster-api/issues/580
+                if upgrade {
+                    debug!("Detecting cluster upgrade, ensuring that the legacy resource set is deleted");
+
+                    let client = cluster.client().await?;
+                    // TODO: make sure the "incorrect" resources are deleted if they exist
```
mnaser, line 72: This is the missing work here that I would say is more important, what needs to happen here is that we need to pull all the resources we used to install before via manifest, check if their matchLabels are the correct _new_ ones, if they're not, delete them.. because the next few lines will create the new cloud-provider secret.  we'll need to cover manila csi, cinder csi and cloud controller manager.

mnaser (commented):
added some commentary, i think we're almost there to fix this.

src/driver.rs
```diff
@@ -53,9 +55,35 @@ impl Driver {
...
+                //               the upgrade.
+                //
+                //               https://github.com/vexxhost/magnum-cluster-api/issues/580
+                if upgrade {
+                    debug!("Detecting cluster upgrade, ensuring that the legacy resource set is deleted");
+
+                    let client = cluster.client().await?;
+                    let daemonsets = client.list_resources(
+                        Api::<DaemonSet>::namespaced(client.clone(), "kube-system"),
+                        ListParams::default().labels("k8s-app=openstack-cloud-controller-manager"),
```
mnaser, line 75: This arbitrarily deletes things all the time, no?  Wouldn't we want to make sure that we only delete the ones that using the old labels and keep the ones with the new labels (aka noop)?

In addition, we need to do this for the Cinder CSI and Manila CSI as well.

mnaser (commented):
@okozachenko1203 we're almost there.

src/driver.rs
okozachenko1203: I defined addons for cinder-csi and manila-csi.
Now the next step is to offload these CSIs from legacy CRS and create dedicated CRSs for them separately and use addons.

mnaser: @okozachenko1203 I think we are adding extra scope here, I am not opposed to doing this, however, the issue is currently upgrades are broken because of this.

We really need to ship an upgrade fix, so I think we should keep this narrowly focused to upgrades, and then we can do the follow up for Manila and Cinder.

okozachenko1203: @mnaser got it

mnaser:
In order to simplify reviews, I've refactored the status change fields into https://github.com/vexxhost/magnum-cluster-api/pull/711

mnaser:
@okozachenko1203 i have something cleaner here, I made a list of all the old resource names:

## Old OpenStack cloud controller manager resources
- `ds/openstack-cloud-controller-manager`

## Old Cinder CSI resources
- `deploy/csi-cinder-controllerplugin`
- `ds/csi-cinder-nodeplugin`

## Old Manila CSI resources
- `sts/openstack-manila-csi-controllerplugin`
- `ds/openstack-manila-csi-nodeplugin`

Well, luckily for us, most of the them had their names changed except the CCM, but we can identify that by label as you've caught:

## Resources to delete
- `deploy/csi-cinder-controllerplugin`
- `ds/csi-cinder-nodeplugin`
- `ds/openstack-cloud-controller-manager` (label `k8s-app=openstack-cloud-controller-manager`)
- `ds/openstack-manila-csi-nodeplugin`
- `sts/openstack-manila-csi-controllerplugin`

--------------

So, if we just delete those, which I've adjusted the code to do, we should be good to go.

mnaser:
@okozachenko1203 ok i think we're in a good spot, can you run upgrade tests on your side please?

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/580

PR #651, opened 2025-04-30
add csi addons

okozachenko1203:


okozachenko1203:
@mnaser  i can create the cluster successfully in my local lab.
now ci fails at devstack installation. i think it would be good to lock devstack version in CI.

okozachenko1203:
recheck

okozachenko1203:
recheck

okozachenko1203:
depends-on: https://github.com/vexxhost/magnum-cluster-api/pull/638

okozachenko1203:
recheck

magnum_cluster_api/charts/patches/openstack-manila-csi/001-add-ca-certificate-and-set-namespace.patch
mnaser, line 1: Can we split this into two patches and push these upstream as well?

magnum_cluster_api/resources.py
mnaser, line 1: This seems to only cover Cinder which I've completed but it does not cover generating the manifests for Manila and moving it into here.

magnum_cluster_api/tests/unit/cmd/test_image_loader.py
```diff
@@ -52,7 +52,7 @@ def test__get_calico_images():
 def test__get_infra_images():
     manifests_path = pkg_resources.resource_filename("magnum_cluster_api", "manifests")
 
-    for csi in ["cinder", "manila", "nfs"]:
```
mnaser, line 55: I assume `manila` should be dropped here.

src/addons/manila-csi.rs
mnaser, line 1: Can you revise this to be the same format as how the cinder CSI was built out?

src/driver.rs
mnaser, line 1: This will need the Manila CSI to be added as well.

mnaser (changes requested):

magnum_cluster_api/charts/patches/openstack-manila-csi/001-add-ca-certificate-and-set-namespace.patch
okozachenko1203: https://github.com/kubernetes/cloud-provider-openstack/pull/2889
https://github.com/kubernetes/cloud-provider-openstack/pull/2890

okozachenko1203:
@mnaser sonobuoy fails with SynchronizedBeforeSuite in most jobs. Do you have any idea on this error?

mnaser:
> @mnaser i can see that sonobuoy fails because cinder-csi-plugin is not ready with `Failed to GetOpenStackProvider : No suitable endpoint could be found in the service catalog.` error. Because i cannot reproduce this issue in local for any versions of k8s, I assume this is because devstack installation by gh action has some difference in openstack service catalog. csi requires volumev3. I need to jump into runner to dig but seems the runner is stuck again. Could you release some runners?

@okozachenko1203 I think the issue is here:

https://github.com/kubernetes/cloud-provider-openstack/issues/2884

We may need to workaround by manually creating the service / endpoint in this case.  I would really hate to pin to an older version of DevStack which we might eventually forget, lol

mnaser:
> @mnaser i can see that sonobuoy fails because cinder-csi-plugin is not ready with `Failed to GetOpenStackProvider : No suitable endpoint could be found in the service catalog.` error. Because i cannot reproduce this issue in local for any versions of k8s, I assume this is because devstack installation by gh action has some difference in openstack service catalog. csi requires volumev3. I need to jump into runner to dig but seems the runner is stuck again. Could you release some runners?

Also I believe I fixed the issue for the workers (for real this time).

okozachenko1203:
> > @mnaser i can see that sonobuoy fails because cinder-csi-plugin is not ready with `Failed to GetOpenStackProvider : No suitable endpoint could be found in the service catalog.` error. Because i cannot reproduce this issue in local for any versions of k8s, I assume this is because devstack installation by gh action has some difference in openstack service catalog. csi requires volumev3. I need to jump into runner to dig but seems the runner is stuck again. Could you release some runners?
> 
> @okozachenko1203 I think the issue is here:
> 
> [kubernetes/cloud-provider-openstack#2884](https://github.com/kubernetes/cloud-provider-openstack/issues/2884)
> 
> We may need to workaround by manually creating the service / endpoint in this case. I would really hate to pin to an older version of DevStack which we might eventually forget, lol

got it. lets see how this workaround works https://github.com/vexxhost/magnum-cluster-api/pull/703

magnum_cluster_api/resources.py
```diff
@@ -351,61 +389,14 @@ def get_object(self) -> dict:
...
-            }
-
         if manila.is_enabled(self.cluster):
             share_types = osc.manila().share_types.list()
             share_network_id = self.cluster.labels.get("manila_csi_share_network_id")
             data = {
                 **data,
+                **magnum_cluster_api.Driver.get_manila_csi_cluster_resource_secret_data(
+                    self.cluster
+                ),
```
mnaser, line 399: You should move this into the other secret which we do automatic reconciling of instead of this one, cause no change will happen here.

magnum_cluster_api/resources.py
```diff
@@ -351,61 +389,14 @@ def get_object(self) -> dict:
...
                 **{
                     "manila-csi-secret.yaml": yaml.dump(
```
mnaser, line 401: This should also be moved.

src/addons/manila_csi.rs
```diff
@@ -0,0 +1,326 @@
...
+pub struct CSIValues {
+    csimanila: CSIComponent,
+    nodeplugin: CSINodePlugin,
+    controllerplugin: CSIControllerPlugin,
+}
+
+#[derive(Debug, Deserialize, PartialEq, Serialize)]
+struct CSIComponent {
+    image: ImageDetails,
+}
```
mnaser, line 22: Is this the same CSIComponent used in Cinder?  If so, let's create a `csi.rs` module and extract it to in order to refactor and not duplicate stuff :)

mnaser (changes requested):
We need to delete this:

https://github.com/vexxhost/magnum-cluster-api/tree/main/magnum_cluster_api/manifests/manila-csi

mnaser (approved):

Issue #479, opened 2025-01-15
Security group rules for worker nodes when octavia_provider is OVN


gabriel-samfira:
Hi folks,

According to https://cluster-api-openstack.sigs.k8s.io/clusteropenstack/configuration.html#security-groups, when OVN is used as a load balancer, the following spec needs to be used:

```yaml
managedSecurityGroups:
  workerNodesSecurityGroupRules:
  - remoteIPPrefix: 0.0.0.0/0
    direction: ingress
    etherType: IPv4
    name: Node Port (TCP, anywhere)
    portRangeMin: 30000
    portRangeMax: 32767
    protocol: tcp
  - remoteIPPrefix: 0.0.0.0/0
    direction: ingress
    etherType: IPv4
    name: Node Port (UDP, anywhere)
    portRangeMin: 30000
    portRangeMax: 32767
    protocol: udp
```

Right now, there is no connectivity when using OVN as a load balancer.

gabriel-samfira:
gah. This seems to only have been made available in the `main` branch of the openstack provider on the 8th of January :sweat_smile:.

mnaser:
@gabriel-samfira I'm okay with making this the default if you want to push up a patch.

krisstoffe:
Hello,

I encountered the same "problem".
Here is the modification I made :

```
diff --git a/src/features/mod.rs b/src/features/mod.rs
index f110f70..9a47f41 100644
--- a/src/features/mod.rs
+++ b/src/features/mod.rs
@@ -35,6 +35,7 @@ use crate::cluster_api::{
         OpenStackClusterTemplate, OpenStackClusterTemplateSpec, OpenStackClusterTemplateTemplate,
         OpenStackClusterTemplateTemplateSpec, OpenStackClusterTemplateTemplateSpecIdentityRef,
         OpenStackClusterTemplateTemplateSpecManagedSecurityGroups,
+        OpenStackClusterTemplateTemplateSpecManagedSecurityGroupsAllNodesSecurityGroupRules,
     },
     openstackmachinetemplates::{
         OpenStackMachineTemplate, OpenStackMachineTemplateSpec, OpenStackMachineTemplateTemplate,
@@ -314,6 +315,28 @@ pub static OPENSTACK_CLUSTER_TEMPLATE: LazyLock<OpenStackClusterTemplate> =
                     managed_security_groups: Some(
                         OpenStackClusterTemplateTemplateSpecManagedSecurityGroups {
[...]

PR #654, opened 2025-05-03
enable traffic from anywhere to node port services on all nodes

krisstoffe:
related to #479


mnaser:
recheck

krisstoffe:
I spotted the now useless ``` ..Default::default()``` in the PR thanks to the checks. I will make an update.

PR #703, opened 2025-05-15
ci: create volumev3 service and endpoint

okozachenko1203:
https://github.com/kubernetes/cloud-provider-openstack/issues/2884

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,13 +147,24 @@ jobs:
             [nova_client]
             api_version = 2.15
           enabled_services: -s-account,-s-container,-s-object,-s-proxy,openstack-cli-server,octavia,o-api,o-hk,o-da
-      - run: ./hack/run-integration-tests.sh
+      - run: |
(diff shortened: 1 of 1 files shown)


mnaser:
@okozachenko1203 I'm quite confused to be honest, since the most recent push to main actually passed CI just fine.

https://github.com/vexxhost/magnum-cluster-api/actions/runs/15033665842

So is it possible this is not necessary?

okozachenko1203:
> @okozachenko1203 I'm quite confused to be honest, since the most recent push to main actually passed CI just fine.
> 
> https://github.com/vexxhost/magnum-cluster-api/actions/runs/15033665842
> 
> So is it possible this is not necessary?

there is no volumev3 service exists so `cinder.is_enabled` returns false. So cinder-csi is never installed in those CIs :D

okozachenko1203:
but in csi-addon PR, we changed this code to `block-storage` so it installs cinder-csi and fails
`common.is_enabled(cluster, "cinder_csi_enabled", "volumev3")`

Issue #740, opened 2025-06-23
Migrate `manila.csi.openstack.org` StorageClass to using `nfs.manila.csi.openstack.org`


mnaser:
The old storage classes were created with a CSIDriver that did not have a prefix of the protocol type, where as the new ones are using NFS and the correct prefix.

We should simply make it so that the old ones are dropped and replaced by the new ones as part of this.

PR #748, opened 2025-07-02
Fix csi manila upgrade case

ricolin:
Fix manila storageclass upgrade
`provisioner` is immutable. for PV. PV in cluster that is built with storageclass that changes `provisioner` will not be able to upgrade to new storageclass. As we use same storageclass name, old cluster is forcing to upgrade to new storageclass and fail as it ask for `provisioner` change in PV.

Fix cinder_csi_enabled and manila_csi_enabled type
These two type are passing as string in cluster.labels not bool.

Relate to https://github.com/vexxhost/magnum-cluster-api/issues/740

### DIFF
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -338,7 +338,7 @@ class CloudProviderClusterResourcesSecret(ClusterBase):
                                 "allowVolumeExpansion": True,
                                 "kind": objects.StorageClass.kind,
                                 "metadata": {
-                                    "name": "share-%s"
+                                    "name": "share-nfs-%s"
                                     % utils.convert_to_rfc1123(st.name),
                                 },
                                 "provisioner": "nfs.manila.csi.openstack.org",
--- a/src/addons/cinder_csi.rs
+++ b/src/addons/cinder_csi.rs
@@ -200,8 +200,8 @@ impl ClusterAddon for Addon {
         Self { cluster }
     }
 
-    fn enabled(&self) -> bool {
-        self.cluster.labels.cinder_csi_enabled
+    fn enabled(&self) ->  bool {
+        self.cluster.labels.cinder_csi_enabled.parse::<bool>().expect("failed to fetch cluster label")
     }
 
     fn secret_name(&self) -> Result<String, ClusterError> {
--- a/src/addons/manila_csi.rs
+++ b/src/addons/manila_csi.rs
@@ -128,8 +128,8 @@ impl ClusterAddon for Addon {
         Self { cluster }
     }
 
-    fn enabled(&self) -> bool {
-        self.cluster.labels.manila_csi_enabled
+    fn enabled(&self) ->  bool {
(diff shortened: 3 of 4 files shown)


magnum_cluster_api/resources.py
```diff
@@ -338,7 +338,7 @@ def get_object(self) -> dict:
                                 "allowVolumeExpansion": True,
                                 "kind": objects.StorageClass.kind,
                                 "metadata": {
-                                    "name": "share-%s"
+                                    "name": "share-nfs-%s"
```
okozachenko1203, line 341: The good thing is that we can delete SC(StorageClass) in use without any issue.
So I'd prefer to check current SC's provisioner field and delete if that is old one(without proto prefix). And create a new one. And all those operations in the code, not manual.

ricolin: The upgrade is stuck with existing PVs that already have "provisioner" defined and which cannot be easily modified.

ricolin: It's in PVC on 
```
kind: PersistentVolumeClaim
metadata:
  annotations:
    ...
    volume.beta.kubernetes.io/storage-provisioner: manila.csi.openstack.org
    volume.kubernetes.io/storage-provisioner: manila.csi.openstack.org
    ...
```


And here in PV
```
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/provisioned-by: manila.csi.openstack.org
spec:
  ...
  csi:
    ...
    driver: manila.csi.openstack.org
    ...
    volumeAttributes:
      ...
      storage.kubernetes.io/csiProvisionerIdentity: 1722446784123-8081-manila.csi.openstack.org
    ...
```

ricolin:
Also currently possible workaround is to recreate PVC with new storage driver and move PV to bond with new PVC as follow:
```
First, change the delete policy to retain
export target_pv=pvc-7cde8337-d7a7-4ed1-9645-a9983aec3be3
kubectl get pv/$target_pv -o yaml | sed 's/persistentVolumeReclaimPolicy.*/persistentVolumeReclaimPolicy: Retain/g' | kubectl replace -f -
 
recreate pvc and sc
Clear dependency on PV (like attach from pod) to make sure pv can change to status Released
delete pvc manila-pvc like
kubectl delete pvc manila-pvc
recreate sc with "nfs.manila.csi.openstack.org" as provisioner
 
create pvc with
with volumeName: $target_pv
 
remove entire claimRef scope in $target_pv
 
it will re-bond with new pvc
 
After we can edit
metadata.annotations."pv.kubernetes.io/provisioned-by" to nfs.manila.csi.openstack.org
 
as for spec.csi.driver
It can't be changed, but I think it should be okay in this case.
 
Change back the reclaim policy:
kubectl get pv/$target_pv -o yaml | sed 's/persistentVolumeReclaimPolicy.*/persistentVolumeReclaimPolicy: Delete/g' | kubectl replace -f -
 
```

PR #751, opened 2025-07-23
Add ability to set admission control plugins

klindgren-godaddy:
Add support for enabling admission control plugins via the admission_control_list label on a cluster or cluster template. https://docs.openstack.org/magnum/latest/user/#admission-control-list

### DIFF
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -896,6 +896,18 @@ class Cluster(ClusterBase):
             self.pykube_api, namespace=self.namespace
         ).get_or_none(name=self.cluster.stack_id)
 
+    def _get_admission_control_list(self) -> str:
+        """Get admission control list with NodeRestriction as base.
+
+        This matches the behavior of the Heat templates where NodeRestriction
+        is always prepended to any user-provided admission plugins.
+        """
+        user_plugins = self.cluster.labels.get("admission_control_list", "")
+        if user_plugins:
+            return f"NodeRestriction,{user_plugins}"
+        else:
+            return "NodeRestriction"
+
     def get_object(self) -> dict:
         osc = clients.get_openstack_api(self.context)
         default_volume_type = osc.cinder().volume_types.default()
@@ -1201,6 +1213,10 @@ class Cluster(ClusterBase):
                             "value": utils.is_controlplane_different_failure_domain(
                                 cluster=self.cluster
                             ),
+                        },
+                        {
+                            "name": "admissionControlList",
+                            "value": self._get_admission_control_list(),
                         },
                     ],
                 },
--- a/src/features/admission_plugins.rs
+++ b/src/features/admission_plugins.rs
@@ -1,0 +1,207 @@
+use crate::{
+    cluster_api::{
+        clusterclasses::{
+            ClusterClassPatches, ClusterClassPatchesDefinitions,
+            ClusterClassPatchesDefinitionsJsonPatches,
+            ClusterClassPatchesDefinitionsJsonPatchesValueFrom,
+            ClusterClassPatchesDefinitionsSelector,
+            ClusterClassPatchesDefinitionsSelectorMatchResources, ClusterClassVariables,
+            ClusterClassVariablesSchema,
+        },
+        kubeadmcontrolplanetemplates::KubeadmControlPlaneTemplate,
+    },
+    features::{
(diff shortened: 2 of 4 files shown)


magnum_cluster_api/resources.py
```diff
@@ -896,6 +896,18 @@ def get_or_none(self) -> objects.Cluster:
...
+        """Get admission control list with NodeRestriction as base.
+
+        This matches the behavior of the Heat templates where NodeRestriction
+        is always prepended to any user-provided admission plugins.
+        """
+        user_plugins = self.cluster.labels.get("admission_control_list", "")
+        if user_plugins:
+            return f"NodeRestriction,{user_plugins}"
+        else:
+            return "NodeRestriction"
```
mnaser, line 909: Doesn't this code enable NodeRestriction arbitrarily?

src/features/admission_plugins.rs
```diff
@@ -0,0 +1,207 @@
...
+    },
+};
+use cluster_feature_derive::ClusterFeatureValues;
+use kube::CustomResourceExt;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, ClusterFeatureValues)]
+pub struct FeatureValues {
+    #[serde(rename = "admissionControlList")]
+    pub admission_control_list: String,
```
mnaser, line 25: Why don't we make this a list instead?

src/features/admission_plugins.rs
```diff
@@ -0,0 +1,207 @@
...
+                    match_resources: ClusterClassPatchesDefinitionsSelectorMatchResources {
+                        control_plane: Some(true),
+                        ..Default::default()
+                    },
+                },
+                json_patches: vec![ClusterClassPatchesDefinitionsJsonPatches {
+                    op: "add".into(),
+                    path: "/spec/template/spec/kubeadmConfigSpec/clusterConfiguration/apiServer/extraArgs/enable-admission-plugins".into(),
+                    value_from: Some(ClusterClassPatchesDefinitionsJsonPatchesValueFrom {
+                        variable: Some("admissionControlList".into()),
```
mnaser, line 48: and maybe we can join it here after? what do you think?

src/features/admission_plugins.rs
```diff
@@ -0,0 +1,207 @@
...
+    use super::*;
+    use crate::{features::test::TestClusterResources, resources::fixtures::default_values};
+    use maplit::btreemap;
+    use pretty_assertions::assert_eq;
+
+    #[test]
+    fn test_disabled_if_admission_control_list_is_empty() {
+        let feature = Feature {};
+        let mut values = default_values();
+        values.admission_control_list = "".to_string();
```
mnaser, line 74: I think it will never be empty by the code logic.

magnum_cluster_api/resources.py
klindgren-godaddy: The default for the enabled-admission-plugins is already NodeRestriction?  As is shown by my own cluster, but also the in the tests? https://github.com/vexxhost/magnum-cluster-api/blob/48d5f2afc0e9856f14168b5e6b97445240102530/tests/fixtures/kube-apiserver.yaml#L20
`--enable-admission-plugins=NodeRestriction`

Additionally, this maps to what is done in the other magnum drivers where they append the admission_control_list to NodeRestriction:  https://github.com/openstack/magnum/blob/fa1c5ae66d9eda7ec8eabb5113f9e7f9dbb28f35/magnum/drivers/common/templates/kubernetes/fragments/configure-kubernetes-master.sh#L64
`KUBE_ADMISSION_CONTROL="--admission-control=NodeRestriction,${ADMISSION_CONTROL_LIST}"`

okozachenko1203: there is no need to set `NodeRestriction` explicitly as it is set in default

src/features/admission_plugins.rs
okozachenko1203: yeah, @klindgren-godaddy this will fail because of https://github.com/vexxhost/magnum-cluster-api/pull/751/files#diff-b21e2a96624fabac44d6c5ecc34bff0835565999be6a320fe75bd0401d5c8e87R430-R432

mnaser:
approved workflows, let's see how CI feels about this.

mnaser (approved):

PR #834, opened 2025-10-17
feat: add octavia_provider config option with amphora deprecation warning

okozachenko1203:
Add [driver] octavia_provider configuration option (default: amphora) and log warning when legacy amphora provider is detected to encourage migration to amphorav2.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1094,7 +1094,7 @@ checksum = "34080505efa8e45a4b816c349525ebe327ceaa8559756f0356cba97ef3bf7432"
 
 [[package]]
 name = "magnum-cluster-api"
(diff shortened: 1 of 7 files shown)


mnaser (changes requested):
Since 5 years ago, `amphorav2` is supported and it simply will be the same but _without_ the jobboard stuff, so let's just change the default to `amphorav2` and that's it.

okozachenko1203:
> Since 5 years ago, `amphorav2` is supported and it simply will be the same but _without_ the jobboard stuff, so let's just change the default to `amphorav2` and that's it.

ok done

okozachenko1203:
@mnaser can you review again?

PR #842, opened 2025-10-23
ci: fix lint and clippy errors

okozachenko1203:
updated clippy version has strict check for dead code detection


src/cluster_api/clusterclasses.rs
```diff
@@ -212,6 +212,7 @@ pub struct ClusterClassControlPlaneMachineHealthCheck {
 /// creates a new object from the template referenced and hands off remediation of the machine to
 /// a controller that lives outside of Cluster API.
 #[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, JsonSchema)]
+#[allow(dead_code)]
```
mnaser: This code is automatically generated, so it will be overwritten.

Issue #790, opened 2025-09-03
With http_proxy settings configured magnum_cluster_api refuses to connect to kube clusters


klindgren-godaddy:
Problem:
set http_proxy, https_proxy, and no_proxy env variables:
  export no_proxy="<kubecluster ip>"
  export http_proxy="<your http proxy>:3128"
  export HTTPS_PROXY="<your https proxy>:3128"

Attempt to use magnum_cluster_api to connect to the management cluster to build a new workload cluster and you will see an error like:

`2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall [-] Fixed interval looping call 'magnum.service.periodic.ClusterUpdateJob.update_status' failed: RuntimeError: configured proxy http://10.33.124.130:3128/ requires the disabled
      feature "kube/http-proxy"
      2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall Traceback (most recent call last):
      2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall   File "/var/lib/kolla/venv/lib64/python3.11/site-packages/oslo_service/loopingcall.py", line 150, in _run_loop
      2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall     result = func(*self.args, **self.kw)
      2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall   File "/var/lib/kolla/venv/lib64/python3.11/site-packages/magnum/service/periodic.py", line 72, in update_status
      2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall     cdriver = driver.Driver.get_driver_for_cluster(self.ctx, self.cluster)
      2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall   File "/var/lib/kolla/venv/lib64/python3.11/site-packages/magnum/drivers/common/driver.py", line 165, in get_driver_for_cluster
      2025-09-02 16:53:45.847 7 ERROR oslo.service.loopingcall     return cls.get_driver(ct.server_type, ct.cluster_distro, ct.coe,
[...]

PR #844, opened 2025-10-28
Fix empty proxy configuration breaking CAPI validation

PatriiCloud:
Fixes vexxhost/magnum-cluster-api#790

When HTTP/HTTPS proxy is not configured, empty strings returned by generate_systemd_proxy_config() and generate_apt_proxy_config() were being base64-encoded and causing YAML parsing errors in Cluster API.

The issue occurs because:
1. utils.generate_systemd_proxy_config() returns "" when no proxy is set
2. utils.generate_apt_proxy_config() returns "" when no proxy is set
3. These empty strings get base64-encoded in resources.py
4. When CAPI decodes them, empty values break cloud-init YAML parsing
5. Cluster creation fails with validation errors

The fix:
- Use "or '#'" fallback in resources.py when encoding proxy configs
- When proxy functions return "", the fallback "#" is used instead
- "#" is a comment character in both systemd and apt config files
- Base64-encoding "#" is safe and doesn't break YAML parsing
- Clusters with actual proxy configs continue to work normally

Changes:
- magnum_cluster_api/resources.py: Add "or '#'" fallback to proxy config encoding
- magnum_cluster_api/tests/unit/test_resources.py: Add comprehensive unit tests

Test coverage includes:
1. Clusters without proxy configuration (fallback to "#")
2. Clusters with proxy configuration (actual config encoded)
3. Critical test ensuring empty strings are never encoded

This is a minimal, safe fix that:
- Prevents cluster creation failures
- Has no impact on existing proxy-enabled clusters
- Uses a harmless comment character as fallback
- Is compatible with all configuration file formats

Related issue: https://github.com/vexxhost/magnum-cluster-api/issues/790

### DIFF
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -1121,13 +1121,13 @@ class Cluster(ClusterBase):
                         {
                             "name": "systemdProxyConfig",
                             "value": base64.encode_as_text(
-                                utils.generate_systemd_proxy_config(self.cluster)
+                                utils.generate_systemd_proxy_config(self.cluster) or "#"
                             ),
                         },
                         {
                             "name": "aptProxyConfig",
                             "value": base64.encode_as_text(
-                                utils.generate_apt_proxy_config(self.cluster)
+                                utils.generate_apt_proxy_config(self.cluster) or "#"
                             ),
                         },
                         {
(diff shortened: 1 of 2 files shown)


mnaser:
@jaszil what version of cluster api are you running if you dont mind me asking?

mvngne:
I'm having the same issues. It would be great if this could be fixed soon.

PR #846, opened 2025-11-12
feat: refresh service account token automatically

okozachenko1203:


mnaser (changes requested):
This change needs tests

mnaser (approved):

PR #847, opened 2025-11-13
fix: recreate openstackcluster to update immutable fields

okozachenko1203:
The commit https://github.com/vexxhost/magnum-cluster-api/commit/004315737efe5eb1c4b1d067c2a5533788fcbd61 changes openstackcluster spec which is immutable.
Upgrade existing clusters to a newer version introducing this change results upgrade failed because of attempts updating immutable fields in openstackcluster.

This PR recreates openstackcluster CR without destroying underlying resources.

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -120,7 +120,7 @@ class BaseDriver(driver.Driver):
             cluster,
             skip_auto_scaling_release=True,
         )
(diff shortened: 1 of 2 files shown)


mnaser (changes requested):
Please review internal ticket about how to handle this case.

mnaser:
replaced by https://github.com/vexxhost/magnum-cluster-api/pull/850

PR #849, opened 2025-11-13
Add `address-sort-order` to cluster config

### LINKED ISSUE #442: fr: support setting kubelet --node-ip using workers' primary NIC
We are using secondary networking in some Kubernetes clusters via Multus. Unfortunately, when we attach additional Neutron ports to k8s worker VMs (or if bare metal nodes have additional NICs), the IP addresses for these are automatically added to the node's list of InternalIPs. As the secondary network is not universally available (for example it's not used by controller nodes), attaching these interfaces can break connectivity to pods' log output and the ability to run `kubectl exec`.

It would be helpful if `--node-ip` could be set for the kubelet at deploy time based on the IP it is allocated. I can't immediately think of a reason this would need to be restricted to this use case, so it may be possible to do so universally.

samdbmg:
Fixes #442 

Adds `address-sort-order` to the Cloud Controller Manager config, set to the CIDR of the node subnet - this feature was [added in v1.26.0](https://github.com/kubernetes/cloud-provider-openstack/commit/effc737).

This causes that address to sort before any secondary NICs on the node, avoiding the problem of secondary Neutron ports being used when trying to get log output.

I've tested this in @andrewbonney's deployment and it solves the problem in #442 as a result (although I've not yet been able to run the unit tests locally, so I'm hoping CI has me covered!)

### DIFF
--- a/magnum_cluster_api/tests/unit/test_utils.py
+++ b/magnum_cluster_api/tests/unit/test_utils.py
@@ -112,6 +112,9 @@ class TestGenerateCloudControllerManagerConfig:
             lb-provider=amphora
             lb-method=ROUND_ROBIN
             create-monitor=True
+
+            [Networking]
+            address-sort-order=10.0.0.0/24
             """
         )
 
@@ -138,6 +141,9 @@ class TestGenerateCloudControllerManagerConfig:
             lb-provider=amphora
             lb-method=ROUND_ROBIN
             create-monitor=True
+
+            [Networking]
+            address-sort-order=10.0.0.0/24
             """
         )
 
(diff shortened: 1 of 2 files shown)


mnaser:
@samdbmg thanks for the awesome change, I do have a small concern:

What will happen in the scenario where someone uses an existing fixed network, in this case, I believe `fixed_subnet_cidr` will be set to the 10.x but won't actually be that, will the deployment fail or how does Kubernetes react if the CIDR does not match any interfaces on the system?

(on a side note, we probably should really have a test for that, but slowly and surely)

PR #850, opened 2025-11-14
fix: make apiServerLoadbalancer sub-fields optional

okozachenko1203:


src/features/mod.rs
```diff
@@ -102,10 +191,26 @@ impl ClusterClassVariablesSchemaExt for ClusterClassVariablesSchema {
...
+            if let Some(schema) = map.get("schema") {
+                schema.clone()
+            } else {
+                json_schema
+            }
+        } else {
+            json_schema
+        };
+
+        let json_schema_str = serde_json::to_string(&schema_value).unwrap();
```
mnaser: Why is this needed?  Isn't the correct schema technically that it can be both a string and null?

magnum_cluster_api/resources.py
```diff
@@ -995,12 +995,17 @@ def get_object(self) -> dict:
         api_server_load_balancer = {
             "enabled": self.cluster.master_lb_enabled,
             "provider": self.cluster.labels.get("octavia_provider", "amphora"),
-            "availabilityZone": self.cluster.labels.get(
-                "api_server_lb_availability_zone", ""
-            ),
-            "flavor": self.cluster.labels.get("api_server_lb_flavor", ""),
```
mnaser: Could we technically replace this with .get(..) without a empty string and that will result in `None` which will go in as null (or None in Rust) and would be just fine and clean?

mnaser (changes requested):

magnum_cluster_api/resources.py
okozachenko1203: I tried `.get()` without default values at the first time. But it added null for those fields and admission webhook rejected this request. So instead of omitting the fields, it added `null` for the fields which require `string`.

src/features/mod.rs
mnaser: I tested this code:

```
use schemars::{JsonSchema, SchemaGenerator};
use serde::{Deserialize, Serialize};
use typed_builder::TypedBuilder;

#[derive(Clone, Serialize, Deserialize, JsonSchema, TypedBuilder)]
pub struct APIServerLoadBalancerConfig {
    pub enabled: bool,

    pub provider: String,

    #[serde(default, skip_serializing_if = "str::is_empty")]
    #[builder(default)]
    pub flavor: String,

    #[serde(default, skip_serializing_if = "str::is_empty", rename = "availabilityZone")]
    #[builder(default)]
    pub availability_zone: String,
}

fn main() {
    let gen = SchemaGenerator::default();
    let schema = gen.into_root_schema_for::<APIServerLoadBalancerConfig>();

    println!("{}", serde_json::to_string_pretty(&schema).unwrap());
}
```

And this renders the following:

```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "APIServerLoadBalancerConfig",
  "type": "object",
  "required": [
    "enabled",
    "provider"
  ],
  "properties": {
    "availabilityZone": {
      "type": "string"
    },
    "enabled": {
      "type": "boolean"
    },
    "flavor": {
      "type": "string"
    },
    "provider": {
      "type": "string"
    }
  }
}
```

This is the correct and accurate form, because technically, those values cannot be null in theory, they are optional, but they cannot be null.

magnum_cluster_api/resources.py
mnaser: I actually wonder now that I have corrected this to become a string again, if it will send an empty string or simply omit...

mnaser: I think it'll stil lbreak, so this is still good.

mnaser (approved):

PR #870, opened 2026-01-16
feat: add support for Cilium HubbleUI

yaguangtang:
Add cilium_hubble_ui_enabled cluster label to enable Cilium's Hubble UI for network observability. When enabled, both the Hubble Relay and Hubble UI components are deployed, allowing users to visualize network flows and service dependencies in their clusters.

Closes: MCAPI-9


fitbeard:
To make it work with private image registry image loader list should be updated and url for UI and other images should be able to overrite.

PR #889, opened 2026-03-03
feat: support kubelet server certificate bootstrap

okozachenko1203:
serverTLSBootstrap enables server certificate bootstrap. Instead of self signing a serving certificate, the Kubelet will request a certificate from the 'certificates.k8s.io' API.

CSR approval is handled by the [kubelet-csr-approver](https://github.com/postfinance/kubelet-csr-approver) operator, deployed as a ClusterResourceSet.

## Changes

- **Native kubeadm patches mechanism** — uses kubeadm's built-in `patches.directory` support instead of fragile `postKubeadmCommands` shell scripts. A strategic merge patch file (`kubeletconfiguration0+strategic.json`) is injected via CAPI's `files` field, and `initConfiguration.patches.directory` / `joinConfiguration.patches.directory` are set so kubeadm applies `serverTLSBootstrap: true` to `KubeletConfiguration` *before* kubelet starts — no post-hoc config editing or kubelet restart needed.

- **API server configuration** — sets `kubelet-certificate-authority: /etc/kubernetes/pki/ca.crt` so the API server verifies kubelet serving certificates using the cluster CA.

- **Worker node coverage** — applies the same kubelet config patch to both control-plane nodes (via `KubeadmControlPlaneTemplate`) and worker nodes (via `KubeadmConfigTemplate`).

- **External CSR approver** — deploys `kubelet-csr-approver` as a Helm chart via `ClusterResourceSet` (in `LegacyClusterResourcesSecret`), configured with `bypassDnsResolution` and `providerIpPrefixes` from the cluster's subnet CIDR.

- **Cluster label** — feature is enabled via the `kubelet_serving_tls_enabled` cluster label, exposed as the `enableKubeletServingTLS` topology variable (optional, defaults to `false`).

---

🔒 GitHub Advanced Security automatically protects Copilot coding agent pull requests. You can protect all pull requests by enabling Advanced Security for your repositories. [Learn more about Advanced Security.](https://gh.io/cca-advanced-security)


okozachenko1203:
it works as expected. now i can see all CSRs are created and approved.

mnaser (changes requested):
Thanks for getting this to a working state and for moving the kubelet config change to kubeadm's native patches mechanism. I think we should make one architectural change before merging: please move the `kubelet-csr-approver` installation out of the Python `LegacyClusterResourcesSecret` path and into the Rust addon pattern used by the cloud provider, CSI, and Cilium addons.

Concretely, that should mean a Rust `ClusterAddon` module that owns the typed values, is enabled from `kubelet_serving_tls_enabled`, renders the vendored chart through the Rust Helm helper, and contributes the manifest through the Rust addon Secret/ClusterResourceSet path rather than `helm.TemplateReleaseCommand` in Python.

While doing that, please also cover the consequences this currently exposes:

- Preserve `container_infra_prefix` image mirroring for `ghcr.io/postfinance/kubelet-csr-approver`.
- Do not apply the chart's Helm test Job via ClusterResourceSet; the current `helm template` output includes the `busybox` test hook as a real child-cluster Job.
- Remove the leftover worker `post_kubeadm_commands: ["echo PLACEHOLDER"]` addition now that this feature uses kubeadm patches instead of post-kubeadm commands.

The kubeadm patching and `providerIpPrefixes`-based CSR approver configuration look like the right direction; this request is mostly to align the addon installation with the newer Rust-based addon format before merge.

PR #964, opened 2026-04-11
chore: unset the template octaiva_provider label

yaguangtang:


mnaser:
i'm 50/50 on this because it slows down CI a bit, let's see by how much.. i think we also disable the `amphorav2` deployment.. so it might not even upload an image or build one.

PR #973, opened 2026-04-14
docs: add CNI upgrade guide for Cilium and Calico

larainema:
## Summary

Add comprehensive CNI upgrade documentation for Magnum-provisioned Kubernetes clusters, covering both Cilium and Calico upgrade paths.

### Documentation (`docs/user/cni-upgrades.md`)
- **Cilium upgrade** with two options:
  - **Option A**: Adopt as Helm release (recommended for ongoing management)
  - **Option B**: Template + `kubectl apply --server-side` (simpler, matches Magnum model)
- **Calico upgrade** with two methods:
  - **Method 1**: Manifest-based (matches Magnum deployment model)
  - **Method 2**: Migrate to Tigera Operator with Helm (recommended)
- Post-upgrade verification steps
- Troubleshooting guide for common issues

### Helm values files
- `cilium-upgrade-values.yaml` — Minimal, tested values preserving Magnum defaults
- `calico-upgrade-values.yaml` — Tigera Operator values for Method 2

## Tested upgrade paths

All paths were validated end-to-end on a DevStack environment:
All paths were validated end-to-end on a DevStack environment:--All paths were valin AAll paths were validated end-to-end on a DevStack environment:--All paths were valin AAll paths were validated end-to-end on a Dev.15.3 All paths were � v1.All paths were validated end-to-end on a De| CalicAll paths were validated end-to-end � vAll paths were validated end-to-end on a DevStack environment:--All paths3.31All paths were validated end-to-end on a DevStack environment:--All paths were valin AAll pyeAll paths were validated end-to-es do NAll paths were validated end-to-end on a DevStack environmentlectorAll paths were validated end-to-end on a DevStack environment:--et/hubble-server-certs` lack labels — must be explicitly annotated or `helm upgrade --install` fails with "invalid ownership metadata"
3. **Calico Method 2**: Tigera Operator auto-migrates pods from `kube-system` to `calico-system` namespace
4. **Calico Method 2**: Version jumps work (v3.27.4 → v3.31.4 direct) despite docs suggesting matching versions only
5. **Cilium values**: Only essential overrides needed — Helm chart provides correct defaults for sub-components (envoy, certgen, nodeinit, etc.)

### DIFF
--- a/docs/user/cni-upgrades.md
+++ b/docs/user/cni-upgrades.md
@@ -1,0 +1,569 @@
+# CNI Upgrade Guide
+
+This guide provides step-by-step instructions for upgrading the Container
+Network Interface (CNI) plugin within your Magnum-provisioned Kubernetes
+cluster. It covers both **Cilium** and **Calico** upgrade procedures.
+
+!!! warning "Read before proceeding"
+
+    CNI upgrades affect pod-to-pod networking. While both Cilium and Calico
+    support rolling upgrades with minimal disruption, it is strongly recommended
+    to:
+
+    - **Test the upgrade in a non-production cluster first.**
+    - **Ensure you have a recent etcd backup** (if applicable).
+    - **Drain and cordon nodes** if you want to be extra cautious.
+    - **Review the upstream release notes** for breaking changes specific to
+      your version jump.
+
+## Prerequisites
+
+Before starting any CNI upgrade, ensure you have the following:
+
+1. **`kubectl` access** to the target Magnum cluster with cluster-admin
+   privileges.
+2. **Helm v3.10+** installed locally.
+3. **Network connectivity** to pull updated container images (or a local
+   registry mirror configured via the `container_infra_prefix` label).
+4. Verify the current CNI and version running in your cluster:
+
+    ```bash
+    # For Cilium
+    kubectl get pods -n kube-system -l app.kubernetes.io/name=cilium -o wide
+    cilium_pod=$(kubectl get pods -n kube-system -l app.kubernetes.io/name=cilium-agent -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || \
+                 kubectl get pods -n kube-system -l k8s-app=cilium -o jsonpath='{.items[0].metadata.name}')
+    kubectl exec -n kube-system "$cilium_pod" -- cilium version
+
+    # For Calico
+    kubectl get pods -n kube-system -l k8s-app=calico-node -o wide
+    kubectl get pods -n kube-system -l k8s-app=calico-node \
+      -o jsonpath='{.items[0].spec.containers[0].image}'
+    ```
+
+---
+
+## Cilium Upgrade
+
+Cilium is deployed in Magnum clusters via a Helm chart (vendored at version
+**1.15.3**). The recommended upgrade path is to use Helm directly against the
+running cluster.
+
+### Supported upgrade path
+
+Cilium supports upgrading **one minor version at a time** (e.g. 1.15 → 1.16).
+Skipping minor versions is **not** supported. Always consult the
+[Cilium Upgrade Guide](https://docs.cilium.io/en/stable/operations/upgrade/)
+for your target version.
+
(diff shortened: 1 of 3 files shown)


docs/user/cni-upgrades.md
```diff
@@ -0,0 +1,555 @@
...
+    reapply the original Cilium version. To prevent this:
+
+    1. **Do not run `openstack coe cluster update`** on CNI-related labels
+       unless the platform operator has updated the `cilium_tag` to your
+       target version.
+    2. If you adopted Cilium as a Helm release (Option A), the Helm release
+       will take precedence on conflicts, but new ClusterResourceSet
+       reconciliations may still create duplicate resources.
+    3. Coordinate with your platform operator to update the `cilium_tag` label
+       in the cluster template to match your upgraded version.
```
jessica-hofmeister: The way I'm reading this is the platform operator (the people providing the environment for end-users to use) has the responsibility to ensure the cilium tag is set to the version they want end-users to be running - so when an end-user does a coe cluster upgrade they get that version. Is that right?
So it would become a maintenance task for the platform operators to update the cilium tag periodically.

docs/user/cni-upgrades.md
```diff
@@ -0,0 +1,555 @@
...
+    The Magnum Cluster API driver manages Cilium via a **ClusterResourceSet**.
+    If the cluster is updated or reconciled by Magnum, it may attempt to
+    reapply the original Cilium version. To prevent this:
+
+    1. **Do not run `openstack coe cluster update`** on CNI-related labels
+       unless the platform operator has updated the `cilium_tag` to your
+       target version.
+    2. If you adopted Cilium as a Helm release (Option A), the Helm release
+       will take precedence on conflicts, but new ClusterResourceSet
+       reconciliations may still create duplicate resources.
```
jessica-hofmeister: If duplicate resources do occur, what steps should be taken to remove them or otherwise resolve the situation?

docs/user/cni-upgrades.md
larainema: Yes, that is correct. Today the platform operator sets `cilium_tag` (or `calico_tag`) on the cluster template, but that tag only takes effect at cluster **creation** time. Magnum does not re-apply the CNI during `coe cluster upgrade` (see [issue #919](https://github.com/vexxhost/magnum-cluster-api/issues/919) for the technical details on why).

So for this manual upgrade guide, the coordination step is: after the end-user upgrades CNI on their cluster, the platform operator should update the template `cilium_tag` to match, so that a future `coe cluster upgrade` does not produce a conflict.

Longer term, we are evaluating automated CNI upgrades (tracked in #919) which would make this transparent — the CNI would upgrade alongside Kubernetes during `coe cluster upgrade`.

docs/user/cni-upgrades.md
larainema: Good catch — I've updated the wording. The legacy ClusterResourceSet actually uses `ApplyOnce` strategy, which means it only applies CNI manifests at cluster creation and does **not** reapply them during subsequent updates or reconciliations. So duplicates are unlikely during normal operation.

I've updated both sections (Step 7 and the bottom "Preventing reversion" section) to clarify this, and added resolution steps in case a conflict ever does occur:

```bash
# Find and delete the legacy CRS to stop it from competing
kubectl get clusterresourceset -n magnum-system
kubectl delete clusterresourceset <cluster-uuid> -n magnum-system
```

docs/user/cni-upgrades.md
nguyenhuukhoi: Could I ask when I use autoscale then new nodes with use the new cni version?

larainema: Yes — new nodes added by the cluster autoscaler will use the upgraded CNI version. The CNI DaemonSet (Cilium or Calico) runs on every node, so when a new node joins the cluster, the DaemonSet controller automatically schedules the CNI pod with the already-upgraded image. No additional action is needed for autoscaled nodes.

nguyenhuukhoi: Thank you, :)

PR #486, opened 2025-01-31
feat: set disableAPIServerFloatingIP as optional boolean

okozachenko1203:
related upstream issue: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2408

PR #983, opened 2026-04-19
fix: preserve disableAPIServerFloatingIP on upgraded clusters

mnaser:
## Summary

Clusters created by magnum-cluster-api versions **older than v0.25.x** had `disableAPIServerFloatingIP` unconditionally patched onto the `OpenStackCluster` spec. [PR #486](https://github.com/vexxhost/magnum-cluster-api/pull/486) (v0.25.0) later guarded that patch with an `enabledIf` so the field is only set when the user actually asks to disable the API-server floating IP.

When such an old cluster is reconciled by a newer magnum-cluster-api, the generated `OpenStackCluster` spec no longer contains the field, but the stored spec still does. CAPO's admission webhook treats any change to `DisableAPIServerFloatingIP` — including `nil ↔ value` — as an immutability violation (it falls back to `reflect.DeepEqual` after zeroing out the fields it explicitly allows to mutate; see [`pkg/webhooks/openstackcluster_webhook.go`](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/release-0.11/pkg/webhooks/openstackcluster_webhook.go)). The topology update is rejected and the upgrade gets stuck.

This is what was observed by @yaguangtang on a customer environment where a cluster cloned from `magnum-v0.31.2` has `disableAPIServerFloatingIP: false` in its spec but a fresh install of 0.31.2 does not.

## Is this environment-specific?

**No** — it affects *every* environment that has clusters originally created by any magnum-cluster-api version that predates PR #486 (v0.25.0). The field was persisted at create time and remains on the `OpenStackCluster` forever because CAPO refuses to let it be removed.

## The fix

The feature is split into two topology variables:

| variable | role |
| --- | --- |
| `disableAPIServerFloatingIP` | the value (unchanged) |
| `disableAPIServerFloatingIPManaged` | gates whether the ClusterClass patch fires |

The Python driver seeds both from the user's `master_lb_floating_ip_enabled` label. The Rust `resolve_immutable_fields` driver additionally flips `disableAPIServerFloatingIPManaged` to `true` — and overrides the value — whenever the existing `OpenStackCluster` spec already has the field set. The patch therefore fires with the preserved value, the generated spec matches the stored spec, and CAPO's webhook is satisfied.

Fresh clusters where the user wants the default (floating IP enabled) remain unaffected: both variables are `false`, the patch is skipped and the field stays absent — which is *also* critical, because adding the field on an `UPDATE` of an existing cluster that never had it would itself trip the same immutability check.

## Tests

* `immutable_fields::tests::disable_api_server_floating_ip_preservation` covers the full matrix (field present/absent, value preservation, managed-flag propagation, caller-supplied overrides).
* `features::disable_api_server_floating_ip::tests::test_patches_if_managed_even_when_value_is_false` asserts that the patch fires and carries `false` through when the resolver has flagged the field as managed.

All 142 Rust unit tests pass locally.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,29 @@
+# Cluster API driver for Magnum
+
+## Checklist for work
+
+When completing any work, make sure that the following is completed:
+
+- All potential regressions are accounted for
+- "pre-commit" has been ran on the codebase.
+
+## Regressions
+
+### Immutable-field preservation
+
+- `src/immutable_fields.rs` has a resolver that reads the existing
+  `OpenStackCluster` spec and overrides topology variables to preserve fields
+  that CAPO's admission webhook considers immutable. Adding a new
+  `FieldMapping` is only half the fix.
+- The Python `Cluster.get_object` builder in `magnum_cluster_api/resources.py`
+  **must** source every resolver-owned key from the `variables` dict returned
+  by `self.rust_driver.resolve_immutable_fields(...)`, the way
+  `apiServerLoadBalancer` does at `resources.py:1048`. Recomputing the value
+  from labels or `utils.*` helpers silently discards the resolver output.
+- This applies to companion/gating flags too (e.g.
+  `disableAPIServerFloatingIPManaged`), not just the primary field.
+- The `mock_rust_driver` fixture in `magnum_cluster_api/tests/unit/conftest.py`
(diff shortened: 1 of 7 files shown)


okozachenko1203 (approved):

mnaser:
recheck

PR #867, opened 2025-12-15
fix: set default value for kube_tag in rust driver

okozachenko1203:
fix https://github.com/vexxhost/magnum-cluster-api/issues/821

Issue #821, opened 2025-10-07
bug: Both cluster creation and deletion are stuck when kube_tag label is not set


okozachenko1203:
### **Description**

When the `kube_tag` label is missing in both the **Cluster Template** and **Cluster** definition, cluster creation and deletion operations fail and end up stuck in `*_FAILED` status.

This happens because the driver code assumes that `kube_tag` is always present and attempts to access it directly without checking for its existence, leading to a `KeyError`.

---

### **Steps to Reproduce**

1. Create a cluster template **without** setting the `kube_tag` label.
2. Create a cluster using that template.
3. Observe that the cluster creation fails with `CREATE_FAILED`.
4. Attempt to delete the same cluster.
5. The deletion also fails with `DELETE_FAILED`.

---

### **Expected Behavior**

* The system should **validate** the presence of the `kube_tag` label during cluster creation.
* If missing, it should raise a **user-friendly validation error** (e.g., `"kube_tag label is required"`) instead of continuing.
* Deletion should not hang or fail due to the missing label.

---

### **Actual Behavior**

Both creation and deletion operations fail due to an unhandled exception:

```

2025-10-06 23:44:32.179 9 ERROR oslo_messaging.rpc.server KeyError: 'kube_tag'
2025-10-06 23:44:32.179 9 ERROR oslo_messaging.rpc.server 
2025-10-06 23:44:32.179 9 ERROR oslo_messaging.rpc.server The above exception was the direct cause of the following exception:
2025-10-06 23:44:32.179 9 ERROR oslo_messaging.rpc.server 
2025-10-06 23:44:32.179 9 ERROR oslo_messaging.rpc.server Traceback (most recent call last):
2025-10-06 23:44:32.179 9 ERROR oslo_messaging.rpc.server   File "/var/lib/openstack/lib/python3.10/site-packages/oslo_messaging/rpc/server.py", line 172, in _process_incoming
2025-10-06 23:44:32.179 9 ERROR oslo_messaging.rpc.server     res = self.dispatcher.dispatch(message)
2025-10-06 23:44:32.179 9 ERROR oslo_messaging.rpc.server   File "/var/lib/openstack/lib/python3.10/site-packages/oslo_messaging/rpc/dispatcher.py", line 309, in dispatch
2025-10-06 23:44:32.179 9 ERROR oslo_messaging.rpc.server     return self._do_dispatch(endpoint, method, ctxt, args)
[...]

PR #995, opened 2026-04-24
fix(#821): require kube_tag label and fail fast with actionable error

larainema:
## Why

The `kube_tag` label drives kubelet/kubeadm bootstrap and **must** match the Kubernetes version baked into the cluster image. Today `get_kube_tag()` silently falls back to a hard-coded `v1.25.3` when the label is missing, which has two failure modes:

1. The default cannot possibly be correct for an arbitrary cluster image, so it produces clusters that fail to bootstrap with no clear hint as to why.
2. When the label is omitted entirely, downstream code paths (including the Rust driver during cluster delete) raise a bare `KeyError: 'kube_tag'` that is opaque to operators. See #821 for the original traceback.

## What

* New `MissingClusterLabel(magnum.common.exception.Invalid)` exception with an actionable message that names the missing label and shows how to set it.
* `get_kube_tag()` now raises `MissingClusterLabel` instead of returning a misleading default.
* `validate_cluster()` calls `get_kube_tag()` first so cluster create is rejected up-front with a clear 400-class error rather than failing partway through bootstrap.
* Unit tests for both `get_kube_tag()` and `validate_cluster()`.

## Alternative to #867

#867 keeps the silent-default behaviour but moves the literal `v1.25.3` into a Rust constant exported as `magnum_cluster_api.DEFAULT_KUBE_TAG`.

This PR takes the opposite approach: refuse to guess. The label is required and the operator gets a clear error explaining what to sThis PR takes the opposite approach: refuse to guess. The label is required and the operator gets a clear error explaining what to sThis PR takes the opposite approach: refuse to guess. The label is required and t exThisng manual workaround (set `kube_tag` in the DB and re-run delete) still applies for legacy clusters. A follow-up could make the Rust delete path tolerant of a missing label.

Closes-Bug: #821

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -71,9 +71,9 @@ class BaseDriver(driver.Driver):
         cluster.stack_id = utils.generate_cluster_api_name(self.k8s_api)
         cluster.save()
 
-        self.rust_driver.create_cluster(cluster)
-
         utils.validate_cluster(context, cluster)
+
+        self.rust_driver.create_cluster(cluster)
 
         return self._create_cluster(context, cluster)
 
@@ -441,7 +441,11 @@ class BaseDriver(driver.Driver):
         #               https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/990
         utils.delete_loadbalancers(context, cluster)
 
-        self.rust_driver.delete_cluster(cluster)
+        # NOTE(okozachenko): If kube_tag is missing, the cluster never passed
+        #                     validation during create, so the Rust-managed
+        #                     ClusterResourceSets were never created.
+        if cluster.labels.get("kube_tag"):
+            self.rust_driver.delete_cluster(cluster)
         resources.Cluster(
             context,
             self.kube_client,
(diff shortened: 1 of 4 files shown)


okozachenko1203:
Tested this PR on a devstack environment. Two issues found during testing:

**1. Create cluster still hits a raw `TypeError` before [validate_cluster] runs**

When creating a cluster without `kube_tag` in labels, the error is:

```
TypeError: failed to extract field Cluster.labels
KeyError: 'kube_tag'
```

This happens because in `create_cluster` (driver.py), `self.rust_driver.create_cluster(cluster)` is called **before** `utils.validate_cluster(context, cluster)`. The Rust `ClusterLabels` struct requires `kube_tag`, so PyO3 deserialization fails with a raw `KeyError` instead of the user-friendly `MissingClusterLabel` exception.

**Fix:** Move `utils.validate_cluster(context, cluster)` before `self.rust_driver.create_cluster(cluster)` in `create_cluster`.

**2. Delete cluster fails for clusters that were created without `kube_tag`**

After creation fails (now with a proper error), attempting to delete the cluster also fails with the same `TypeError: failed to extract field Cluster.labels` / `KeyError: 'kube_tag'`. This is because `delete_cluster` calls `self.rust_driver.delete_cluster(cluster)`, which also deserializes the cluster through PyO3 and hits the same missing `kube_tag` issue.

Since [validate_cluster] now runs before [rust_driver.create_cluster], if `kube_tag` was missing, the Rust-managed `ClusterResourceSet`s were never created in the first place — so there's nothing for the Rust driver to delete.

**Fix:** Guard `self.rust_driver.delete_cluster(cluster)` with `if cluster.labels.get("kube_tag")` so deletion succeeds for clusters that never passed validation.

PR #925, opened 2026-04-02
ci: drop pkg_resources and enable Zuul unit test job

mnaser:
- [x] Drop `pkg_resources` and use `importlib.resources.files()`
- [x] Add `magnum-cluster-api-tox-unit` job definition to `zuul.d/jobs.yaml`
- [x] Add `magnum-cluster-api-tox-unit` to check and gate pipelines in `zuul.d/project.yaml`

PR #1001, opened 2026-04-27
fix: migrate off pkg_resources to importlib.resources

ricolin:
## Summary

Replace all uses of `pkg_resources` in `magnum_cluster_api` (and tests) with `importlib.resources`, eliminating a critical packaging-driven failure that prevents `magnum-conductor` from loading the Cluster API driver.

## Problem

When `magnum_cluster_api.driver` is imported, it transitively imports `magnum_cluster_api.resources`, which does:

```python
import pkg_resources
...
pkg_resources.resource_filename("magnum_cluster_api", "charts")
```

`pkg_resources.resource_filename` calls `get_provider()` → `get_distribution()` → `Distribution.activate()`, which resolves `Requires-Dist` for every transitive dependency. Unlike `importlib.metadata`, `pkg_resources` does **not** apply PEP 503 name normalization when matching distribution names against on-disk `*.dist-info` directories.

Recent `oslo.cache` releases (≥ 3.10.2, uploaded 2025-07-25) ship under a PEP 625 normalized wheel filename:

```
oslo_cache-3.10.2-py3-none-any.whl   (was oslo.cache-3.10.1-py3-none-any.whl)
```

PEP 427 then mandates the dist-info directory match the wheel filename, so on disk we get `oslo_cache-3.10.2.dist-info/`. `pkg_resources` keys this distribution as `oslo-cache` (dash), but `keystonemiddleware`'s `Requires-Dist` entry is `oslo.cache` (dot). The two don't match, and `pkg_resources` raises:

```
DistributionNotFound: The 'oslo.cache>=...' distribution was not found
and is required by keystonemiddleware
```

This propagates up through Magnum's driver entry-point loader, so **all cluster operations fail** with:

```
Cluster type (vm, ubuntu, kubernetes) not supported
```

…and async deletes silently never run, because the conductor cannot instantiate the driver.

## Fix

Replace `pkg_resources.resource_filename(pkg, sub)` with `importlib.resources.files(pkg) / sub` everywhere in `magnum_cluster_api/resources.py` and the tests under `magnum_cluster_api/tests/unit/`. `importlib.resources.files` is stdlib since Python 3.9 and uses PEP 503 normalization, so it resolves `oslo.cache` and `oslo_cache-*.dist-info` to the same canonical key.

After this change `magnum_cluster_api` no longer imports `pkg_resources` at all, breaking the failure chain at its source.

## Verification

Reproduced the failure in a magnum-conductor pod (image `ghcr.io/vexxhost/magnum:2025.1`):

```
$ python -c "import importlib; importlib.import_module('magnum_cluster_api.driver')"
...
File ".../magnum_cluster_api/resources.py", line 23, in <module>
    import pkg_resources
File ".../pkg_resources/__init__.py", line 870, in _resolve_dist
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'oslo.cache>=1.26.0' distribution
was not found and is required by keystonemiddleware
```

With this patch, the import succeeds and entry-point loading proceeds normally.

## Affected Releases

The bug surfaces in any container image that bundles `oslo.cache>=3.10.2` together with this codebase (`vexxhost/magnum:2025.1`, `2025.2`, `main`). Older images pinned to `oslo.cache<=3.10.1` are unaffected because the wheel filename was still dotted there. This patch fixes the issue at the source so it is independent of which `oslo.cache` version is installed.

## Notes

- `importlib.resources.files()` returns a `Traversable`, which is path-like; `str(...)` is used at the call sites that pass it into `os.path.join()`.
- No behavior change for the resource lookups themselves — same files, same paths.
- Drops a deprecated dependency on `pkg_resources`, which setuptools is removing.

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c3c2f2c1c457b00c10c4848d6f5491db3b629df # v2.18.0
+        uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
         with:
           egress-policy: audit
 
@@ -42,7 +42,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c3c2f2c1c457b00c10c4848d6f5491db3b629df # v2.18.0
+        uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
         with:
           egress-policy: audit
 
@@ -70,7 +70,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c3c2f2c1c457b00c10c4848d6f5491db3b629df # v2.18.0
+        uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
         with:
           egress-policy: audit
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
         target: [x86_64, aarch64]
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c3c2f2c1c457b00c10c4848d6f5491db3b629df # v2.18.0
+        uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
         with:
           egress-policy: audit
 
@@ -78,7 +78,7 @@ jobs:
             python-version: "3.12"
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@6c3c2f2c1c457b00c10c4848d6f5491db3b629df # v2.18.0
+        uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
         with:
           egress-policy: audit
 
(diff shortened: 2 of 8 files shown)


ricolin:
@copilot apply changes based on the comments in [this thread](https://github.com/vexxhost/magnum-cluster-api/pull/1001#pullrequestreview-4180018666)

ricolin:
@copilot  update this PR and rebase on main. and review again

ricolin:
```
 Fix verified end-to-end

Test: Took the affected production image (ghcr.io/vexxhost/magnum:2025.1@sha256:31e5dd39…, with oslo_cache-3.10.2.dist-info/
and magnum-cluster-api 0.35.0), overlaid only resources.py with the PR #1001 patch applied to that exact 0.35.0 file, and
re-ran Magnum's driver loading.

┌──────────────────────────────┬───────────────────────────────────────┬──────────────────────────────────────────────────────┐
│                              │ Before fix                            │ After fix                                            │
├──────────────────────────────┼───────────────────────────────────────┼──────────────────────────────────────────────────────┤
│ magnum_cluster_api.driver    │ ❌ DistributionNotFound:              │ ✅ OK                                                │
│ import                       │ oslo.cache>=1.26.0                    │                                                      │
├──────────────────────────────┼───────────────────────────────────────┼──────────────────────────────────────────────────────┤
│ pkg_resources loaded         │ ❌ Yes (and crashes)                  │ ✅ No (never imported)                               │
├──────────────────────────────┼───────────────────────────────────────┼──────────────────────────────────────────────────────┤
│ Magnum entry points loaded   │ 0                                     │ 5                                                    │
├──────────────────────────────┼───────────────────────────────────────┼──────────────────────────────────────────────────────┤
│ Driver definition_map        │ empty                                 │ all 5 tuples, including ('vm', 'ubuntu',             │
│ populated                    │                                       │ 'kubernetes')                                        │
├──────────────────────────────┼───────────────────────────────────────┼──────────────────────────────────────────────────────┤
│ CLUSTER_CLASS_VERSION        │ (crashes)                             │ 0.35.0                                               │
└──────────────────────────────┴───────────────────────────────────────┴──────────────────────────────────────────────────────┘
```

ricolin:
duplicated with https://github.com/vexxhost/magnum-cluster-api/pull/925

PR #1003, opened 2026-04-27
ci: smoke-test driver import and entry-point loading

ricolin:
## Summary

The existing `coinstall` job verifies that `magnum-cluster-api` *installs* cleanly against each OpenStack upper-constraints set, but it never imports the package or exercises Magnum's driver entry-point loader. Failures that occur at import time are therefore invisible to CI.

## Why this matters

A real production regression slipped through unnoticed because of this blind spot:

`oslo.cache 3.10.2` (uploaded to PyPI on 2025-07-25, pinned by the `stable/2025.1` upper-constraints) ships under a [PEP 625](https://peps.python.org/pep-0625/) normalized wheel filename — `oslo_cache-3.10.2-py3-none-any.whl` instead of `oslo.cache-*`. [PEP 427](https://peps.python.org/pep-0427/) then mandates the on-disk dist-info directory match the wheel filename, so the install layout becomes `oslo_cache-3.10.2.dist-info/`.

`magnum_cluster_api/resources.py` (pre-#1001) calls `pkg_resources.require("magnum_cluster_api")` at module import time. Legacy `pkg_resources` does not apply [PEP 503](https://peps.python.org/pep-0503/) name normalization across `.`/`_`/`-` when matching distributions, so `keystonemiddleware`'s `oslo.cache>=...` Requires-Dist entry fails to resolve against `oslo_cache-*.dist-info/` and `pkg_resources` raises `DistributionNotFound`.

The driver entry-point then fails to load and Magnum responds to every cluster operation with the misleading:

```
Cluster type (vm, ubuntu, kubernetes) not supported
```

The constraint files used by the existing matrix already include the broken `oslo.cache` versions:

| OpenStack branch | `oslo.cache` constraint |
|---|---|
| `2024.1` | `===3.7.0` (dotted) |
| `2024.2` | `===3.8.0` (dotted) |
| **`2025.1`** | **`===3.10.2` (PEP 625 underscore)** |
| `master` | `===4.1.1` (PEP 625 underscore) |

…so the bug is sitting on disk during CI for the 2025.1 + master cells, but no CI step imports anything to trigger it.

## Change

Add an explicit smoke step to the `coinstall` job that:

1. Installs `magnum` and `keystonemiddleware` so the driver entry-point loader and its real consumer dependency tree are present.
2. Imports `magnum_cluster_api.driver` — must not raise.
3. Loads Magnum's driver entry points via `magnum.drivers.common.driver.Driver.load_entry_points()` and asserts at least one `cluster_api` driver is registered.
4. Asserts that `pkg_resources` was *not* pulled into `sys.modules`. This is the regression guard: if anything reintroduces the legacy resolver (the same class of bug fixed by #1001), every PEP 625-named wheel in the matrix will trip it.

Runs across the full `(openstack-version × python-version)` matrix so the same protection applies to every supported OpenStack branch — including `master`, where future PEP 625 transitions of other dependencies will land first.

## Relationship to #1001

This PR is intentionally orthogonal to #1001:
- **#1001** fixes the actual bug by removing all `pkg_resources` usage from `magnum_cluster_api`.
- **This PR** ensures the bug — and any equivalent future regression — fails CI loudly instead of shipping silently to production.

Drafted to allow review of the test approach before unblocking. Happy to widen / narrow the assertions as needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,42 @@ jobs:
       - run: |
           uv venv
           uv pip install --constraint https://releases.openstack.org/constraints/upper/${{ matrix.openstack-version }} ./*.whl
+      - name: Smoke-test driver import and entry-point loading
+        run: |
+          uv pip install --constraint https://releases.openstack.org/constraints/upper/${{ matrix.openstack-version }} \
+            magnum keystonemiddleware
+          .venv/bin/python - <<'PY'
+          import sys
+          import magnum_cluster_api.driver  # must not raise
+          from magnum.drivers.common import driver
+
+          # Magnum entry points must load
+          eps = list(driver.Driver.load_entry_points())
+          ep_names = sorted(ep.name for ep, _ in eps)
+          assert any("cluster_api" in n for n in ep_names), ep_names
+
+          # Driver.definition_map must be populated — this is what
+          # instantiates each driver class and fails in production when
+          # pkg_resources cannot resolve a PEP 625 oslo_cache-*.dist-info.
+          definition_map = driver.Driver.get_drivers()
+          assert definition_map, "Driver.get_drivers() returned empty definition_map"
+          assert ("vm", "ubuntu", "kubernetes") in definition_map, (
+              "expected (vm, ubuntu, kubernetes) cluster_api driver in "
+              "definition_map, got: %s" % sorted(definition_map)
+          )
+
+          # Regression guard: pkg_resources must never be imported. It
+          # does not apply PEP 503 name normalization across .|_|-, which
+          # is what breaks against PEP 625-named wheels (oslo_cache-*).
+          assert "pkg_resources" not in sys.modules, (
+              "pkg_resources must not be loaded — it breaks on PEP 625 "
+              "wheels (e.g. oslo_cache-*.dist-info)"
+          )
+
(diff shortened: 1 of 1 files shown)


ricolin:
## CI status: blocked on #925

The new `coinstall` smoke-test in this PR is intentionally strict — it asserts that importing `magnum_cluster_api` (and `magnum_cluster_api.resources`) does **not** pull in `pkg_resources`, which is exactly the regression that let the oslo_cache PEP 625 bug slip through.

Against current `main`, `magnum_cluster_api/resources.py` still does `import pkg_resources` at module load (line 23, plus three `pkg_resources.resource_filename(...)` calls), so the assertion fails on every coinstall job by design.

**#925 is the prerequisite** — it drops `pkg_resources` in favour of `importlib.resources.files()` in `resources.py` and the unit tests. It's already open and all-green.

### Why both PRs (not just #925)
- **#925** = the fix + a Zuul `tox-unit` job (single environment).
- **#1003** = GH Actions smoke test running across the **entire `coinstall` matrix** (zed → master × py3.10/3.11/3.12), which is the exact surface where the original oslo_cache PEP 625 regression was invisible. Zuul's single tox env would not have caught it.

### Suggested merge order
1. Land #925 (drops `pkg_resources`).
2. Re-run CI on this PR — the smoke-test should go green automatically and start providing ongoing matrix-wide regression protection.

No code change needed here; happy to rebase once #925 lands.

PR #1010, opened 2026-05-02
feat: add label kubelet_extra_args to enable pass args to the kubelet

### LINKED ISSUE #442: fr: support setting kubelet --node-ip using workers' primary NIC

yaguangtang:
Pass extra command-line flags to the kubelet on every control-plane and worker node in the cluster.
fixes #442

### DIFF
--- a/docs/user/labels.md
+++ b/docs/user/labels.md
@@ -231,6 +231,27 @@ is often accomplished by deploying a driver on each node.
 
    Default value: `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305`
 
+* `kubelet_extra_args`
+
+   Pass extra command-line flags to the kubelet on every control-plane and
+   worker node in the cluster.  These are merged into the
+   `kubeletExtraArgs` map of the underlying `KubeadmControlPlaneTemplate`
+   and `KubeadmConfigTemplate`, in addition to the defaults that the
+   driver always sets (`cloud-provider=external` and `tls-cipher-suites`).
+
+   The label value is a **semicolon-separated** list of `key=value` pairs.
+   Semicolon (rather than comma) is used so that values which themselves
+   contain commas — such as kubelet's `--system-reserved` flag — pass
+   through unchanged.  Keys are kubelet flag names without the leading
+   `--`.  Whitespace around pairs is stripped, and entries without `=` or
+   with an empty key are silently ignored.
+
+   Example:
+
+       kubelet_extra_args=max-pods=150;system-reserved=cpu=100m,memory=128Mi;eviction-hard=memory.available<100Mi
+
+   Default value: `""` (no extra arguments)
+
 * `kube_tag`
 
    The version of Kubernetes to use.
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -1229,6 +1229,10 @@ class Cluster(ClusterBase):
                             ),
                         },
                         {
+                            "name": "kubeletExtraArgs",
+                            "value": utils.get_kubelet_extra_args(self.cluster),
+                        },
+                        {
                             "name": "apiServerSANs",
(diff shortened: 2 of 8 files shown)


yaguangtang:
```
openstack coe cluster template show k8s-v1.34.3-template
+-----------------------+------------------------------------------------------------------------------------------------------------------------------+
| Field                 | Value                                                                                                                        |
+-----------------------+------------------------------------------------------------------------------------------------------------------------------+
| insecure_registry     | -                                                                                                                            |
| labels                | {'kube_tag': 'v1.34.3', 'kubelet_extra_args': 'node-labels=environment=production', 'master_lb_floating_ip_enabled': 'True'} |
| updated_at            | 2026-05-02T11:57:02+00:00                                                                                                    |
| floating_ip_enabled   | True                                                                                                                         |
| fixed_subnet          | -                                                                                                                            |
| master_flavor_id      | m1.medium                                                                                                                    |
| uuid                  | fe17dd0a-d862-4536-96ef-43fd2829330c                                                                                         |
| no_proxy              | -                                                                                                                            |
| https_proxy           | -                                                                                                                            |
| tls_disabled          | False                                                                                                                        |
| keypair_id            | -                                                                                                                            |
| public                | False                                                                                                                        |
| http_proxy            | -                                                                                                                            |
| docker_volume_size    | -                                                                                                                            |
| server_type           | vm                                                                                                                           |
| external_network_id   | public                                                                                                                       |
| cluster_distro        | ubuntu                                                                                                                       |
| image_id              | 7e0fd7f6-dcfb-462e-a3bb-57e2539d923c                                                                                         |
| volume_driver         | -                                                                                                                            |
| registry_enabled      | False                                                                                                                        |
| docker_storage_driver | overlay2                                                                                                                     |
| apiserver_port        | -                                                                                                                            |
| name                  | k8s-v1.34.3-template                                                                                                         |
| created_at            | 2026-05-02T10:40:46+00:00                                                                                                    |
| network_driver        | calico                                                                                                                       |
| fixed_network         | -                                                                                                                            |
| coe                   | kubernetes                                                                                                                   |
| flavor_id             | m1.medium                                                                                                                    |
| master_lb_enabled     | True                                                                                                                         |
| dns_nameserver        | 8.8.8.8                                                                                                                      |
| project_id            | 23f76aeb57b7468988c1c1fe108d9a71                                                                                             |
| hidden                | False                                                                                                                        |
| tags                  | -                                                                                                                            |
+-----------------------+------------------------------------------------------------------------------------------------------------------------------+
```


```
kubectl get node -l environment=production
NAME                                          STATUS   ROLES                  AGE   VERSION
kube-w239s-default-worker-ccvtc-59j4r-9mtms   Ready    worker                 28m   v1.34.3
kube-w239s-vcgr9-mlct9                        Ready    control-plane,master   31m   v1.34.3
```

ricolin (changes requested):
So one of thing I realise is that kubeletExtraArgs method itself is actually not currage in kubelet for cluster wide kubelet configs, and some of it's values are deprecated as mentioned
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#kubelet
I think for node specific, we're good with using kubeletExtraArgs, but for the other changes, we should go with approach like https://github.com/vexxhost/magnum-cluster-api/pull/1013

This should be something documented later


Also just an idea, not tested yet:
one of the approach we can discuss is on top of current works, we can consider to support node ip cidr may like  
--labels kubelet_node_ip_mode=cidr
--labels kubelet_node_ip_mode=10.10.20.0/24
or even the kubelet_node_ip_interface

in this way we don't need to set multiple node group just for set ips for nodes.
we can make this node group overrideable so it can be 10.10.20.0/27 for worker nodes

PR #1011, opened 2026-05-04
fix(driver,sync,utils): converge stuck DELETE_IN_PROGRESS nodegroups under concurrent SSA race

ricolin:
## Summary

Four commits that together harden the magnum-cluster-api reconciler against the
parallel `delete_nodegroup` race that wedges nodegroups in
`DELETE_IN_PROGRESS` indefinitely while their `MachineDeployment` keeps
running.

| # | Commit | What it does |
|---|---|---|
| 1 | `fix(driver): guard reconciler against MachineDeploymentNotFound` | Stop the periodic `update_nodegroups_status` loop from being killed for the lifetime of the conductor pod when a concurrent topology mutation removes an MD between the GET and the spec-equality lookup. |
| 2 | `fix(driver): recover stuck DELETE_IN_PROGRESS nodegroups with live MD` | Add a recovery branch in the reconciler: when an NG is `DELETE_IN_PROGRESS` but its MD is still present (a previous SSA was silently overwritten by a concurrent SSA with the same fieldManager + `force=True`), re-issue the topology removal so the delete eventually converges. |
| 3 | `fix(sync): bump ClusterLock default TTL from 60s to 300s` | The 60s sherlock-KubernetesLock TTL with no heartbeat is shorter than the worst-case driver operation; on expiry, a second conductor can race the original. Bump to 5 minutes as an immediate mitigation; a proper fix needs a heartbeat / auto-renew loop. |
| 4 | `fix(utils): strip resourceVersion from kube_apply_patch body` | pykube serialises the full GET'd object (which carries `metadata.resourceVersion`) into the SSA body. The apiserver then runs its generic optimistic-concurrency check and 409s on every parallel write — this is unrelated to SSA semantics. Strip RV/uid/generation from the body before applying. |

## Race in brief

1. Two `delete_nodegroup` calls run on different conductor pods.
2. Both GET the Cluster CR, each removes a different `machineDeployments` entry, both Apply with the same `fieldManager` (`atmosphere-operator`) + `force=true`.
3. Last writer wins: the earlier removal is silently restored.
4. The DB row for the resurrected NG sits in `DELETE_IN_PROGRESS` forever — there was no recovery branch, and any subsequent `MachineDeploymentNotFound` raised by another NG would crash the whole reconcile loop.

## Verification

End-to-end on an AIO Atmosphere environment running this branch:

- **Lock-protected real-driver `delete_nodegroup` × 10 in parallel**: 4/5 trials clean (10/10 ok, 0 topology leftover); 1 trial transient lock-acquire timeout (lease churn from prior trials, not fix-related — recovery branch heals on next loop).
- **Forced race (10 parallel SSA Applies bypassing ClusterLock)**: 5/5 trials reproduce the original bug (9/10 entries stuck). The recovery branch then converges them.
- **Mixed parallel create + delete**: clean.
- **TTL hold > 60s**: second `delete_nodegroup` waits the full duration, proving the 60→300s bump works as intended.
- **resourceVersion forced-race repro**: 10/10 → 0/10 409s with the strip-RV fix.
- **Unit tests**: 171/180 pass (9 pre-existing baseline failures unrelated to this change).
- **Upgrade safety**: per-commit cross-version analysis + real `create_nodegroup` → `delete_nodegroup` E2E confirms the branch is upgrade-safe with mixed-ownership topology entries from older releases (the original `fieldManager` is preserved, no SSA ownership migration needed).

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -533,6 +533,33 @@ class BaseDriver(driver.Driver):
                 node_group.save()
                 continue
 
+            # NOTE(rlin): If the node group is in `DELETE_IN_PROGRESS` state but
+            #             the `MachineDeployment` is still present, the previous
+            #             delete_nodegroup attempt's Server-Side Apply was lost
+            #             (e.g. a concurrent SSA on the same Cluster resource
+            #             with the same fieldManager silently overwrote the
+            #             topology change). Re-issue the topology removal so
+            #             the delete eventually converges instead of wedging
+            #             forever.
+            if (
+                node_group.status == fields.ClusterStatus.DELETE_IN_PROGRESS
+                and md is not None
+            ):
+                try:
+                    md_index = cluster_resource.get_machine_deployment_index(
+                        node_group.name
+                    )
+                except exceptions.MachineDeploymentNotFound:
+                    # Topology already removed; just wait for CAPI to delete
+                    # the MachineDeployment.
+                    continue
+                else:
+                    del cluster_resource.obj["spec"]["topology"]["workers"][
+                        "machineDeployments"
+                    ][md_index]
+                    utils.kube_apply_patch(cluster_resource)
+                    continue
+
             md_is_running = (
                 md is not None and md.obj.get("status", {}).get("phase") == "Running"
             )
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -558,9 +558,18 @@ class BaseDriver(driver.Driver):
                 },
             )
 
+            # NOTE(rlin): Look up the MachineDeployment spec from the Cluster
+            #             topology. If it's missing (e.g. another concurrent
+            #             operation just removed it from the topology), skip
+            #             the spec-equality checks for this reconcile pass
+            #             rather than crashing the entire periodic loop.
+            try:
+                md_spec = cluster_resource.get_machine_deployment_spec(node_group.name)
+            except exceptions.MachineDeploymentNotFound:
+                continue
+
             # Ensure that the image ID from the spec matches all of the OpenStackMachine objects
             # for this node group
-            md_spec = cluster_resource.get_machine_deployment_spec(node_group.name)
             md_variables = {
                 i["name"]: i["value"] for i in md_spec["variables"]["overrides"]
             }
@@ -587,9 +596,7 @@ class BaseDriver(driver.Driver):
             if (
                 node_group.status == fields.ClusterStatus.UPDATE_IN_PROGRESS
                 and md_is_running
-                and md.equals_spec(
-                    cluster_resource.get_machine_deployment_spec(node_group.name)
-                )
+                and md.equals_spec(md_spec)
                 and image_id_match
                 and flavor_match
             ):
--- a/magnum_cluster_api/sync.py
+++ b/magnum_cluster_api/sync.py
@@ -21,7 +21,20 @@ class ClusterLock(sherlock.KubernetesLock):
     across all of the conductor nodes.
     """
 
-    DEFAULT_EXPIRE: int = 60
+    # NOTE(rlin): The default TTL was 60s, which is too short for several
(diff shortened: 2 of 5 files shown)


ricolin:
@copilot review this PR

ricolin:
## 🟢 End-to-End Verification on Fresh AIO

Overlaid this branch's `utils.py` / `driver.py` / `sync.py` (md5s matching tip `c0130d3`) into a fresh AIO's `magnum-conductor` pod via ConfigMap + StatefulSet patch, then exercised the full Magnum lifecycle.

### Operations Verified

| Operation | Result | Detail |
|---|---|---|
| `cluster-create` (LB + master + worker) | ✅ CREATE_COMPLETE | ~10 min |
| `cluster-resize` (1 → 2 workers) | ✅ UPDATE_COMPLETE | `default-worker` MD: 1/1 → 2/2 |
| `nodegroup-create` (`extra-ng`, 1) | ✅ UPDATE_COMPLETE | new MD `extra-ng` 1/1 |
| **`nodegroup-delete` (`extra-ng`)** | ✅ UPDATE_COMPLETE | **MD removed cleanly — exact path this PR fixes** |

### Fix Behaviour Observed

- No 409/conflict errors during any SSA cycle.
- No stuck `DELETE_IN_PROGRESS` — the recovery branch was not even needed on the clean path.
- `kube_apply_patch` body-stripping: silent/working — no `resourceVersion`/`uid`/`generation` conflicts under normal lifecycle.

### Notes

- Cross-version cluster-upgrade was previously verified on a separate AIO; on this AIO only one kube image (`v1.34.3`) was available, so the upgrade SSA path was exercised here via resize + nodegroup add/remove (which hit the same `kube_apply_patch` code path with topology changes).
- All four PR commits + the unit-test commit are signed-off; CI is green (pre-commit/black, cargo, codeql, full coinstall matrix).

okozachenko1203:
@ricolin good catch, i experienced this and tried to fix. let me test your patch on the lab

magnum_cluster_api/driver.py
```diff
@@ -533,6 +533,33 @@ def update_nodegroups_status(
...
+            ):
+                try:
+                    md_index = cluster_resource.get_machine_deployment_index(
+                        node_group.name
+                    )
+                except exceptions.MachineDeploymentNotFound:
+                    # Topology already removed; just wait for CAPI to delete
+                    # the MachineDeployment.
+                    continue
+                else:
```
okozachenko1203, line 552: I think we can delete the `else` here since the `except` path already does a `continue`.

okozachenko1203:
Tested this PR against the concurrent nodegroup deletion issue.
 
I chose Terraform because it makes concurrent nodegroup deletion easier and more repeatable than running manual `openstack coe nodegroup delete` commands.
 
Test setup:
- Created a Magnum cluster with CAPI/CAPO
- Created 4 additional worker nodegroups
- Initially created the nodegroups without setting Terraform parallelism, but some nodegroups ended in `CREATE_FAILED` due to the resource conflict issue addressed by this PR
- Used `terraform apply -parallelism=1` during creation so I could reproduce the delete issue before the PR without being blocked by create-time conflicts
- Used Terraform with higher parallelism during deletion to trigger concurrent nodegroup deletes
 
Before this PR:
- 3 nodegroups remained stuck in `DELETE_IN_PROGRESS`
- 1 nodegroup ended in `DELETE_FAILED` due to a resource/version conflict
 
After this PR:
- All 4 nodegroups were deleted successfully
- No nodegroups remained stuck in `DELETE_IN_PROGRESS`
- The related CAPI `MachineDeployment` resources were removed as expected

magnum_cluster_api/driver.py
ricolin: the else is for other exceptions other than exceptions.MachineDeploymentNotFound to kinda restore the DB and status so orphan request from race condition will not hurts cluster machines

okozachenko1203:
@ricolin i've tested it and it looks good. Feel free to merge

PR #1014, opened 2026-05-06
feat: support server_type=bm in cluster_api driver (auto-default boot_volume_size=0)

ricolin:
## Problem

Magnum's cluster_api drivers only advertised `server_type=vm`. Baremetal templates using Ironic through Nova failed driver lookup, and baremetal boot also needed safer defaults around root volumes and networking.

## Solution

This PR now carries the `server_type=bm` behavior as three focused pieces:

1. `feat(driver): advertise server_type=bm alongside vm`
   - Adds `server_type=bm` entries for the shipped cluster_api drivers.
2. `feat(resources): default boot_volume_size=0 for server_type=bm`
   - Defaults BM clusters away from Cinder root volumes unless the operator explicitly sets a boot volume size.
3. `fix(pr1014): disable CAPO managed security groups for bm`
   - Adds `disableManagedSecurityGroups` and removes CAPO `managedSecurityGroups` only for `server_type=bm` ClusterClasses.
   - This is Issue 41 from the live OVN BM replay. CAPO managed security groups caused Neutron `PortSecurityAndIPRequiredForSecurityGroups` on the disabled-port-security baremetal VLAN network, before Nova/Ironic server create.

VM clusters keep the existing CAPO managed-security-group behavior.

## Validation

Latest scoped validation after moving Issue 41 into this PR:

- `cargo test managed_security_groups` -> 2 passed
- `cargo test test_convert_values_to_cluster_topology_variables` -> 1 passed
- `uv run pytest magnum_cluster_api/tests/unit/test_resources.py::test_cluster_variables_disable_managed_security_groups_for_bm -q` -> 1 passed, 8 warnings
- `uv run ruff check magnum_cluster_api/resources.py magnum_cluster_api/tests/unit/test_resources.py` -> passed
- `rustfmt --check src/features/managed_security_groups.rs` -> passed

## Review Notes

The managed-security-groups change is deliberately scoped to BM. The live failure came from a BM VLAN network with `port_security_enabled=False`; VM tenant-network behavior should remain unchanged.

### DIFF
--- a/docs/user/labels.md
+++ b/docs/user/labels.md
@@ -17,7 +17,11 @@ specify the volume size and type using the following labels:
 
 :   The size in gigabytes of the boot volume.  If you set this value, it will
     enable boot from volume.
-    **Default value**: Unset
+    **Default value**: Unset for `server_type=vm` clusters; automatically
+    `0` for `server_type=bm` clusters (baremetal nodes boot from local disk
+    via Ironic and cannot use a Cinder root volume). Operators can still
+    override explicitly on `server_type=bm` by setting this label on the
+    cluster_template, cluster, or node_group.
 
 `boot_volume_type`
 
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -732,6 +732,7 @@ class UbuntuDriver(BaseDriver):
     def provides(self):
         return [
             {"server_type": "vm", "os": "ubuntu", "coe": "kubernetes"},
+            {"server_type": "bm", "os": "ubuntu", "coe": "kubernetes"},
         ]
 
 
@@ -740,6 +741,7 @@ class UbuntuFocalDriver(UbuntuDriver):
     def provides(self):
         return [
             {"server_type": "vm", "os": "ubuntu-focal", "coe": "kubernetes"},
+            {"server_type": "bm", "os": "ubuntu-focal", "coe": "kubernetes"},
         ]
 
 
@@ -756,6 +758,7 @@ class FlatcarDriver(BaseDriver):
     def provides(self):
         return [
             {"server_type": "vm", "os": "flatcar", "coe": "kubernetes"},
+            {"server_type": "bm", "os": "flatcar", "coe": "kubernetes"},
         ]
 
 
@@ -764,4 +767,5 @@ class RockyLinuxDriver(BaseDriver):
     def provides(self):
         return [
             {"server_type": "vm", "os": "rockylinux", "coe": "kubernetes"},
+            {"server_type": "bm", "os": "rockylinux", "coe": "kubernetes"},
         ]
--- a/magnum_cluster_api/resources.py
+++ b/magnum_cluster_api/resources.py
@@ -741,7 +741,9 @@ def mutate_machine_deployment(
         boot_volume_size = utils.get_node_group_label_as_int(
             node_group,
             "boot_volume_size",
-            CONF.cinder.default_boot_volume_size,
+            utils.get_default_boot_volume_size(
+                cluster, CONF.cinder.default_boot_volume_size
+            ),
         )
         if boot_volume_size == 0:
             boot_volume_size = flavor.disk
@@ -802,7 +804,10 @@ def mutate_machine_deployment(
                             "size": utils.get_node_group_label_as_int(
                                 node_group,
                                 "boot_volume_size",
-                                CONF.cinder.default_boot_volume_size,
+                                utils.get_default_boot_volume_size(
+                                    cluster,
+                                    CONF.cinder.default_boot_volume_size,
+                                ),
                             ),
                             "type": node_group.labels.get(
                                 "boot_volume_type",
@@ -1106,7 +1111,10 @@ class Cluster(ClusterBase):
                                 "size": utils.get_cluster_label_as_int(
                                     self.cluster,
                                     "boot_volume_size",
-                                    CONF.cinder.default_boot_volume_size,
+                                    utils.get_default_boot_volume_size(
+                                        self.cluster,
+                                        CONF.cinder.default_boot_volume_size,
+                                    ),
                                 ),
                                 "type": self.cluster.labels.get(
                                     "boot_volume_type",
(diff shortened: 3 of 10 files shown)


ricolin:
Result: ✅ End-to-end pass with PR #1014

 - Cluster gpu-test (82c529b9-fd42-4176-8824-dad42ee60eef) reached CREATE_COMPLETE first try.
 - Template used --server-type bm and no boot_volume_size label — exercised PR #1014's driver dispatch + auto-default. Master 
on node-1 (baremetal), worker on node-0 (baremetal-gpu), both Ready with v1.34.3, all kube-system pods Running.

mnaser (changes requested):
Out of curiosity, why do we need `boot_volume_size=0` -- what happens if it's not set?  It seems/feels like it should just work out of the box, no?

ricolin:
> Out of curiosity, why do we need `boot_volume_size=0` -- what happens if it's not set? It seems/feels like it should just work out of the box, no?

@mnaser 
so boot_volume_size defaults to CONF.cinder.default_boot_volume_size, that makes when we try to provide bm for magnum cluster it will failed because machine try to boot the ironic bm from cinder volume.
A workaround is to make sure we alway specifc set boot volume size to 0 from label.
WDYT?

ricolin:
recheck

ricolin:
### Test report (test env redacted)

**Scope:** `--server-type bm` driver dispatch, `boot_volume_size=0` auto-default for BM, autoscaler ephemeral-disk annotation skip-when-0.

**Env:** Atmosphere AIO with 3-node BM emulator (libvirt + OVS PXE bridge), kube image `ubuntu-2204-kube-v1.34.{3,7}-raw`.

| Phase | Test | Result |
|---|---|---|
| C2 | `coe cluster create --server-type bm` (no `boot_volume_size` label) | ✅ PASS — first try, CREATE_COMPLETE / HEALTHY |
| U1 | Rolling upgrade `kube_tag v1.34.3 → v1.34.7` (KCP surge + MD rollout) | ✅ PASS |
| U3 | Worker scale-up `node_count 1 → 2` immediately after U1 | ✅ PASS — UPDATE_COMPLETE / HEALTHY |
| G1 | Fake-GPU device-plugin DaemonSet + demo pod on the upgraded cluster | ✅ PASS |
| Gap | `boot_volume_size=0` auto-default | ✅ Held across create + upgrade + scale |
| Gap | Autoscaler ephemeral-disk annotation skip-when-0 | ✅ Did not regress |
| Gap | BM driver dispatch (no VM-only specifics on BM path) | ✅ Did not regress |

**Overall: 5/5 executed phases PASS.** Round-1 report archived locally with the full reproducer commands, libvirt domain XML, and ovs-vsctl wiring.

Caveat: `Driver().provides` is the public API for "is server_type=bm supported"; no separate helper added.

ricolin:
recheck

ricolin:
## Follow-up: validate `server_type=bm` flavors are Ironic-backed (`c2257ff`)

While integration-testing this PR end-to-end on a baremetal AIO, I hit a silent failure mode that I think this PR should close before merge.

### Repro on this branch (no patch applied)

Cluster template with `server_type=bm` and BM-backed flavors. Override `--master-flavor` with a stock virtual flavor on `cluster create`:

```
$ openstack coe cluster create --cluster-template k8s-bm-gpu \
    --keypair admin-key --master-count 1 --node-count 0 \
    --master-flavor m1.small \
    --labels kube_tag=v1.34.3,octavia_provider=ovn \
    gap-master-vm
Request to create cluster 23fdc519-d4f9-4f36-9e2a-d6980d8a03c6 accepted
```

The cluster reaches `CREATE_IN_PROGRESS` immediately (no validation error) and proceeds to schedule the master. Inspecting the resulting Nova server:

```
$ openstack server show kube-84lu2-rrtgr-r6zkn -c flavor -c hypervisor_hostname -f value
{'name': 'm1.small', 'original_name': 'm1.small', ..., 'extra_specs': {}, ...}
instance        # ← this is the AIO host's KVM hypervisor, not an Ironic node
```

So the master silently lands on a libvirt/KVM hypervisor with `flavor=m1.small`, `extra_specs={}` — none of the `server_type=bm` code paths added by this PR fire (`get_default_boot_volume_size` returns 0 only for BM, the chart's `network_interface=flat` decision downstream depends on the same predicate, etc.). The cluster looks healthy but is not actually testing the BM control plane.

For comparison, the Ironic-backed flavor on the same AIO does have the expected extra-spec:
```
$ openstack flavor show baremetal -f value -c properties
{'capabilities:boot_mode': 'uefi', 'cpu_arch': 'x86_64',
 'resources:CUSTOM_BAREMETAL': '1', 'resources:DISK_GB': '0',
 'resources:MEMORY_MB': '0', 'resources:VCPU': '0', 'trait:CUSTOM_GOLD': 'required'}
```

### Fix in `c2257ff`

Adds `validate_baremetal_flavors()` (and a small `_is_baremetal_flavor()` helper) to `utils.py`, hooked into `validate_cluster()`. A flavor is considered Ironic-backed when its extra-specs contain a `resources:CUSTOM_<NAME>=1` entry, which is the convention Nova's Ironic driver requires for the scheduler to route the request to the Ironic compute service.

The validator is a no-op for `server_type=vm` templates, so the existing virtual-cluster path is unaffected.

### Tests

```
$ pytest magnum_cluster_api/tests/unit/test_utils.py::TestValidateBaremetalFlavors -q
8 passed, 8 warnings in 0.03s
```

Coverage:
- positive / negative / unrelated-extra-specs detection
- fallback to cached `extra_specs` attribute when `get_keys()` is not eagerly resolvable
- no-op for `server_type=vm`
- pass when both flavors are baremetal-backed
- reject when only `master_flavor_id` is virtual
- reject when only `flavor_id` is virtual

The whole `tests/unit/test_utils.py` module: 27 passed, 1 pre-existing failure (`TestGenerateCloudControllerManagerConfig.test_generate_cloud_controller_manager_config` — fixture expects `lb-provider=amphora` but the test environment now returns `amphorav2`; reproduces on `f9a753e` without my changes).

ricolin:
recheck

PR #1015, opened 2026-05-06
feat: extra_cloud_init — operator-controlled files + pre/post-kubeadm commands

ricolin:
## Summary

Adds three new Magnum labels for cloud-init passthrough on every cluster node, working uniformly on **VM and bare-metal** substrates:

- `extra_files` — base64-encoded YAML/JSON list of files to drop on the node.
- `extra_pre_kubeadm_commands` — `;;`-separated shell commands run **before** `kubeadm init/join`.
- `extra_post_kubeadm_commands` — `;;`-separated shell commands run **after** `kubeadm init/join`.

All three honor cluster-level **and** node-group-level labels; node-group entries are appended after cluster entries.

## Driving use case

Apply a custom `netplan` configuration before `kubeadm` runs so multi-NIC / VLAN / bonded layouts come up first:

```bash
PAYLOAD=$(base64 -w0 <<'EOF'
- path: /etc/netplan/99-mcapi.yaml
  permissions: "0600"
  content: |
    network:
      version: 2
      ethernets:
        enp4s0: { dhcp4: true, mtu: 1450 }
EOF
)
openstack coe cluster create ... --labels \
  extra_files=$PAYLOAD,\
  extra_pre_kubeadm_commands="netplan generate;;netplan apply;;sleep 3"
```

The same plumbing supports any operator-injected file or command (CA bundles, sysctl tweaks, etc.) without rebuilding the node image.

## Implementation notes

- **Slot-based ClusterClassPatches.** 10 file slots, 16 pre-command slots, 16 post-command slots.  Each slot has its own `enabledIf: gt (len .extraXxx) <i>` gate and emits an `op: add` against the parent list path.  Per-element appends were chosen over parent-replace because `preKubeadmCommands` is shared with `containerd_config`, `keystone_auth`, and `operating_system` features — replacing it would clobber them.
- **Python helpers** in `magnum_cluster_api/utils.py` validate paths (must be absolute), normalize ownership/permissions, base64-encode raw `content` (or pass through pre-encoded), split commands on `;;`, and enforce per-list caps that mirror the Rust slot counts.
- **`build.rs`** gains an `extract_vec_inner()` helper so a user-defined `Vec<ExtraFile>` gets qualified correctly in the generated `Values` struct.

## Tests

- `cargo test --lib`: 146/146 passing (4 new in `extra_cloud_init`).
[...]

PR #1016, opened 2026-05-08
fix(labels): merge cluster_template labels into sparse cluster.labels

ricolin:
## Summary

Magnum `cluster create --labels` replaces, rather than merges, cluster-template labels. When an operator creates a cluster with sparse custom labels, the stored cluster row can miss labels required by the Rust `ClusterLabels` extractor, such as `kube_tag`.

This PR fills those missing template-derived labels before the Rust create/delete paths and persists the completed label dict on create.

## Fix

- Add `utils.fill_missing_labels_from_template(cluster)` to fill missing labels without overriding user-provided cluster labels.
- Prefer `cluster.labels_skipped` before `cluster.cluster_template.labels`, matching Magnum's stored representation for template-only labels on sparse-label creates.
- Invoke the helper before Rust create/delete paths.
- Persist the filled label dict during create when it changes, so later status/health monitor reads see the completed labels from Magnum storage.
- Tolerate CAPI `Cluster` objects that do not yet have `status.conditions` during early reconciliation.

This PR is rebased on current `main`, which already includes PR #1064.

## Repro

1. Create a cluster template with labels like `kube_tag`, `fixed_subnet_cidr`, and `octavia_provider`.
2. Run `openstack coe cluster create --labels audit_log_enabled=true` while omitting those template labels.
3. The sparse cluster row can later fail Rust label extraction with `KeyError: 'kube_tag'`, especially in status/health monitor or delete paths.

## Validation

```bash
tox -e py310
# 221 passed, 58 warnings

git diff --check origin/main...HEAD
# passed
```

Previously validated on Atmosphere main and stable AIOs with the equivalent sparse-label/status-condition fixes:

- sparse labels filled from `labels_skipped` first, then template labels
- missing CAPI `status.conditions` produced an empty status map instead of `KeyError`
- fresh cluster create passed after updated mCAPI image rollout
- old/default cluster create, mCAPI update, then resize of the same existing cluster passed
- delete path passed with sparse labels present

## Notes

- PR #1064 covers string-valued bool-ish CSI labels and is complementary; it does not replace this sparse-label fallback.
- This supersedes draft PR #1051, which duplicated the same label fallback after unrelated fixes were dropped.
- Originally fixed as part of #1015 commit `1930573` alongside unrelated `extra_cloud_init` fixes; split out here so the label fallback can land independently.

Signed-off-by: Rico Lin <rlin@vexxhost.com>

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -69,6 +69,15 @@ class BaseDriver(driver.Driver):
         cluster.stack_id = utils.generate_cluster_api_name(self.k8s_api)
         cluster.save()
 
+        # See note in delete_cluster: Magnum REPLACES template labels with the
+        # cluster's own --labels list.  Merge any missing keys back in so the
+        # Rust driver and downstream consumers see a complete label dict.
+        original_labels = dict(cluster.labels or {})
+        utils.fill_missing_labels_from_template(cluster)
+        if cluster.labels != original_labels:
+            cluster.labels = dict(cluster.labels)
+            cluster.save()
+
         self.rust_driver.create_cluster(cluster)
 
         utils.validate_cluster(context, cluster)
@@ -219,7 +228,8 @@ class BaseDriver(driver.Driver):
 
             capi_cluster.reload()
             status_map = {
-                c["type"]: c["status"] for c in capi_cluster.obj["status"]["conditions"]
+                c["type"]: c["status"]
+                for c in capi_cluster.obj.get("status", {}).get("conditions", [])
             }
 
             for condition in ("ControlPlaneReady", "InfrastructureReady", "Ready"):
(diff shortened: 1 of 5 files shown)


ricolin:
### Test report (test env redacted)

**Scope:** `utils.fill_missing_labels_from_template` — fill template-only label keys back into `cluster.labels` before invoking the Rust driver, so sparse `--labels` on `coe cluster create` no longer crashes the deserialiser.

**Env:** Atmosphere AIO + 3-node BM emulator, kube image v1.34.7.

| Phase | Test | Result |
|---|---|---|
| Repro (pre-fix) | `coe cluster create` with only `extra_*` labels (no `kube_tag` etc.) → `CREATE_FAILED` with `KeyError: 'kube_tag'` from the Rust `ClusterLabels` deserialiser; orphan then stuck in `DELETE_FAILED` (delete path tripped on the same KeyError) | ❌ as expected (gap-22 reproduced) |
| Post-fix | Same sparse-`--labels` create, with this PR applied | ✅ CREATE_COMPLETE on first try |
| Post-fix | `cluster delete` on a sparse-labels cluster | ✅ DELETE_COMPLETE — no KeyError on the delete path |
| Regression | Round-1 BM cluster lifecycle (create → upgrade → scale) on a fully-labelled cluster | ✅ Did not regress |

This was originally bundled inside #1015's commit `1930573` alongside two `extra_cloud_init` fixes; split out here so the label-merge fallback can land on `main` independently of #1015's larger scope. Same E2E evidence covers both PRs.

ricolin:
recheck

ricolin:
@codex[agent]  review this PR

PR #1018, opened 2026-05-09
fix(keystone_auth): configure API server webhooks with kubeadm patches

ricolin:
## Summary

Configure Keystone token authentication and authorization on kubeadm control planes without rewriting the live static-pod manifest after bootstrap.

- Use native kubeadm patches for both init and join control-plane nodes.
- Add the Keystone authentication and authorization webhook configuration flags.
- Replace the kubeadm default authorization mode with `Node,RBAC,Webhook` while first testing the expected default argument.
- Schedule the host-networked Keystone webhook on both current `control-plane` and legacy `master` labeled nodes.
- Probe the webhook through the same `127.0.0.1:8443` listener used by kube-apiserver.

## Why

The previous feature added webhook flags by rewriting the generated kube-apiserver manifest after kubeadm completed. That could leave duplicate authorization modes, did not naturally cover joined or replacement control-plane nodes, and depended on mutating a live static-pod manifest.

Native kubeadm patches apply the configuration while each control-plane manifest is generated. `Node` and `RBAC` remain ahead of `Webhook`, so normal Kubernetes identities continue to use the built-in authorizers and Keystone authorization is used as the fallback. A JSON Patch `test` operation ensures a future kubeadm argument-order change fails instead of replacing an unrelated command argument.

The webhook uses `hostNetwork` and listens only on loopback. Explicit probe hosts are therefore required because a TCP probe otherwise targets the Pod IP, which is the node IP for a host-networked Pod.

## Validation

- `pre-commit run --all-files`
- `cargo test features::keystone_auth`
- `cargo clippy --lib -- -D warnings`
- `helm lint magnum_cluster_api/charts/k8s-keystone-auth`
- Rendered the chart and verified both probes and the process listener use `127.0.0.1`.

The next end-to-end CI run must still confirm that the workload cloud-controller-manager starts, the control-plane Node receives its OpenStack provider ID, the Keystone DaemonSet becomes Ready without restarts, and a Keystone-token request succeeds.

### DIFF
--- a/magnum_cluster_api/charts/k8s-keystone-auth/templates/daemonset.yaml
+++ b/magnum_cluster_api/charts/k8s-keystone-auth/templates/daemonset.yaml
@@ -57,6 +57,12 @@ spec:
               containerPort: {{ .Values.service.port }}
               hostPort: {{ .Values.service.hostPort }}
               protocol: TCP
+          readinessProbe:
+            tcpSocket:
+              port: https
+          livenessProbe:
+            tcpSocket:
+              port: https
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
--- a/magnum_cluster_api/charts/k8s-keystone-auth/values.yaml
+++ b/magnum_cluster_api/charts/k8s-keystone-auth/values.yaml
@@ -78,8 +78,7 @@ volumeMounts: []
 #   mountPath: "/etc/foo"
 #   readOnly: true
 
-nodeSelector:
-  node-role.kubernetes.io/master: ""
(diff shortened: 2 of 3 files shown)


ricolin:
@codex[agent]  review this  PR

mnaser (changes requested):
If we're going to touch this feature, I'd like us to leverage this:

https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#patches

ricolin:
recheck

ricolin:
recheck

ricolin:
recheck

PR #1019, opened 2026-05-09
feat(occm): gate cloud-controller-manager 'service' controller on octavia_enabled label

ricolin:
## Problem

The OCCM Helm chart ships with `enabledControllers=[cloud-node, cloud-node-lifecycle, route, service]` by default. The `service` controller calls Octavia at controller-manager start-up to warm its load-balancer cache. On clusters where Octavia is not deployed (bare-metal, edge, single-node, dev/test), this call hangs:

* leader-election lease times out → OCCM `CrashLoopBackOff`,
* `cloud-node` never sets `ProviderID` on Nodes → kubelet stays `NotReady`,
* the cluster never reaches a usable state.

Operators currently have no way to ask Magnum/mcapi to skip the Octavia controller — they have to post-edit the Deployment by hand, which doesn't survive reconcile.

## Solution

Add a new mutable Magnum cluster label `octavia_enabled` (default `true` for backwards compatibility) that gates the `service` controller in the OCCM chart values:

```bash
openstack coe cluster template create no-lb \
    --labels octavia_enabled=false,...
```

When `octavia_enabled=false`, OCCM is templated with
`--controllers=cloud-node,cloud-node-lifecycle,route` — no Octavia client is instantiated, OCCM comes up cleanly, `ProviderID` is set on Nodes, and the only thing lost is `Service.type=LoadBalancer` reconciliation, which on a no-Octavia cluster has nothing to reconcile anyway.

Default behaviour is unchanged: existing clusters that don't set the label keep the full controller set.

## Changes

* `src/magnum.rs` — new `ClusterLabels.octavia_enabled` string field + `is_octavia_enabled()` helper.
* `src/addons/cloud_controller_manager.rs` — new `enabledControllers` field on `CloudControllerManagerValues`; populated from `is_octavia_enabled()`.

## Tests

* `cargo test --lib` → 143/143 pass, including a new `test_occm_values_octavia_disabled_omits_service_controller` that asserts the rendered chart values omit the `service` controller when the label is `false`.

## Validation

End-to-end: deployed on a cluster with `octavia_enabled=false`. OCCM Pod reaches `Running 1/1`, no leader-election timeouts, all Nodes report `ProviderID` and become `Ready`.

### DIFF
--- a/src/addons/cloud_controller_manager.rs
+++ b/src/addons/cloud_controller_manager.rs
@@ -79,9 +79,9 @@ impl TryFrom<magnum::Cluster> for CloudControllerManagerValues {
     type Error = ClusterAddonValuesError;
 
     fn try_from(cluster: magnum::Cluster) -> Result<Self, ClusterAddonValuesError> {
-        let values = Self::defaults()?;
+        let default_values = Self::defaults()?;
 
-        let image = DockerImage::parse(&values.image.repository)?;
+        let image = DockerImage::parse(&default_values.image.repository)?;
         let values = Self::builder()
             .image(
                 CloudControllerManagerImageValues::builder()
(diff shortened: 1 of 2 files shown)


ricolin:
recheck

ricolin:
@codex[agent]  review this PR

ricolin:
recheck

PR #1020, opened 2026-05-09
fix(driver): report CAPI Machine failure reasons

ricolin:
## Problem

When a CAPI Machine fails, the parent infrastructure resource can remain ready while the workload cluster is unhealthy. Magnum can therefore expose `UNHEALTHY` with an empty or unhelpful `status_reason`, forcing operators to inspect the management cluster.

The original implementation also reported every false leaf condition. That can duplicate one provider failure through several propagation layers and produce unstable, unbounded output.

## Solution

Use CAPI's canonical readiness summaries and retain compatibility fallbacks:

* Prefer the Cluster v1beta2 `ControlPlaneMachinesReady` and `WorkerMachinesReady` summaries during lifecycle polling.
* For older objects, report one summary per unready Machine: v1beta2 `Ready`, legacy `Ready`, terminal failure fields, then selected legacy readiness conditions.
* Keep output deterministic, cap it at five Machine diagnostics, and truncate individual details to 512 characters.
* Fail open only for expected Kubernetes/API transport errors, while logging them and allowing programming errors to surface.
* Avoid direct Machine lookups and expected not-ready noise while deletion is in progress.
* During ongoing health polling, prefer v1beta2 readiness summaries, preserve existing `<infrastructure-name>.Ready` keys, and add bounded `<infrastructure-name>.Reason` details for unhealthy Machines.

This follows the actual CAPO-to-CAPI condition propagation: provider failures are summarized by the infrastructure-ready and Machine-ready conditions rather than copied as provider-specific condition types onto the CAPI Machine.

## Validation

* `cargo test --lib` — 149 passed.
* `cargo clippy --lib -- -D warnings` — passed.
* `pytest -q magnum_cluster_api/tests/unit/test_driver.py -k MachineConditionsAggregation` — 11 passed.
* `pre-commit run --all-files` — all hooks passed.

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -148,21 +148,21 @@ class BaseDriver(driver.Driver):
                     "cluster.x-k8s.io/cluster-name": cluster.stack_id,
                 },
             )
+
+            messages = []
+            for machine in machines:
+                conditions = (machine.obj.get("status") or {}).get("conditions") or []
+                for cond in conditions:
+                    if cond.get("status") == "False" and cond.get("reason"):
+                        name = machine.obj.get("metadata", {}).get("name", "<unknown>")
+                        msg = cond.get("message") or ""
+                        messages.append(
+                            f"{name}/{cond.get('type')}: {cond.get('reason')}"
+                            + (f" ({msg})" if msg else "")
+                        )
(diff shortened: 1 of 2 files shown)


ricolin:
@codex[agent]  review this PR

PR #1027, opened 2026-05-14
fix: include NFS CSI manifests with Manila CSI

larainema:
## Summary
- include the packaged NFS CSI manifests in the Manila CSI ClusterResourceSet output
- mirror NFS CSI images when `container_infra_prefix` is configured
- add unit coverage for the Manila CSI manifest bundle and mirrored NFS CSI images

## Testing
- `cargo test addons::manila_csi`


ricolin:
recheck

PR #1031, opened 2026-05-19
fix: use direct sdk openstack clients

mnaser:
## Summary

- make `clients.get_openstack_api()` return a real `openstack.connection.Connection` built from Magnum's Keystone session, preserving Magnum's per-service `endpoint_type` and `region_name` config as SDK service options
- use SDK proxies directly for identity, compute, block storage, shared file system, image, load-balancer, and service-catalog operations instead of wrapping Magnum's client objects
- add `openstacksdk` as a direct dependency and cover the SDK connection, Cinder fallback, Manila share-type lookup, Keystone service resource compatibility, and nullable Glance image property paths with unit tests
- normalize nullable `hw_disk_bus` image metadata before writing ClusterClass variables so CAPI never receives JSON `null` for the string-only `hardwareDiskBus` variable

## Testing

- `pre-commit run --all-files`
- `nix-shell -p uv rustc cargo pkg-config openssl kubernetes-helm --run 'uv run pytest magnum_cluster_api/tests/unit/test_clients.py magnum_cluster_api/tests/unit/test_integrations_common.py magnum_cluster_api/tests/unit/test_driver.py::TestDriver::test_create_cluster magnum_cluster_api/tests/unit/test_resources.py'`
- `nix-shell -p uv rustc cargo pkg-config openssl kubernetes-helm --run 'uv run pytest magnum_cluster_api/tests/unit/'` (176 passed, 9 existing unrelated failures)

## Notes

The full local unit suite still has the same unrelated failures seen before this rewrite: update-nodegroup tests need a MachineSet list mock in this dependency set, and one utility test expects `amphora` while current output is `amphorav2`.


mnaser:
recheck

ricolin:
@mnaser  I make some test against real env and find some to fix with in https://github.com/vexxhost/magnum-cluster-api/pull/1035 just seperate to another PR  there so it will be easier for review

mnaser:
> Mohammed Naser (@mnaser) I make some test against real env and find some to fix with in #1035 just seperate to another PR there so it will be easier for review

Thanks for splitting it up, merged.

PR #1033, opened 2026-05-20
fix: migrate OpenStack lookups to openstacksdk

ricolin:
## Summary

- migrate Magnum Cluster API OpenStack resource lookups from legacy service clients to openstacksdk
- preserve Magnum API error behavior for fixed network/subnet and missing nodegroup cases
- replace the remaining delete-time Octavia cleanup path with SDK-native load balancer and floating IP cleanup
- drop unused legacy client dependencies and add a release note

## Validation

Unit tests:

- `env -u KUBECONFIG uv run pytest magnum_cluster_api/tests/unit/ -q`
- Result: `197 passed, 16 warnings`

Live end-to-end validation in a temporary test environment:

- deployed the patched Magnum API, conductor, and cluster-api proxy image
- created a fresh Kubernetes cluster successfully and verified both nodes reached `Ready`
- verified `openstack coe cluster config` and Kubernetes API access
- verified Cinder CSI pods, `block-rbd1` StorageClass, PVC provisioning, volume attach, mount, and write/read with `cinder-ok`
- verified Manila CSI pods started; dynamic Manila PVC provisioning reached the driver/backend path but the backend rejected share-server allocation for the temporary share network
- verified a Kubernetes `Service type=LoadBalancer` with an enabled Octavia provider created an active Octavia load balancer and assigned an ingress address
- verified Service deletion cleaned up the Octavia load balancer
- verified nodegroup create and resize behavior
- verified the missing CAPI `MachineDeployment` nodegroup-delete path by manually deleting the MachineDeployment, then deleting the Magnum nodegroup and observing `HTTP 404`
- verified invalid fixed network and invalid fixed subnet names return expected `HTTP 400` errors without creating failed cluster records
- verified cluster delete cleaned up Magnum/CAPI resources, Nova servers, server groups, Octavia load balancers, Manila test resources, and temporary Neutron resources

Known environment-specific limits from validation:

- the template-default Octavia `ovn` provider could not be used because that provider was not enabled in the test cloud; the positive LoadBalancer test was run with an enabled provider instead
- Manila end-to-end mount could not pass because the Manila backend could not allocate a share server on the temporary share network

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+* migrate OpenStack resource lookups to openstacksdk proxies, including
+  fixed network, fixed subnet, image metadata, volume type, Manila share type,
+  Octavia load balancer, flavor, and server group paths
+* use Magnum's native HTTPNotFound exception for missing node groups instead
+  of importing heatclient
+
 ## [0.36.6](https://github.com/vexxhost/magnum-cluster-api/compare/v0.36.5...v0.36.6) (2026-04-21)
 
 
--- a/magnum_cluster_api/clients.py
+++ b/magnum_cluster_api/clients.py
@@ -14,7 +14,7 @@
 
 import pykube  # type: ignore
 from magnum.common import clients, exception  # type: ignore
-from manilaclient.v2 import client as manilaclient  # type: ignore
+from openstack.shared_file_system.v2 import _proxy as shared_file_system_proxy
 
 
 class OpenStackClients(clients.OpenStackClients):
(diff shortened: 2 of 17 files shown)


ricolin:
OVS live validation update for `a7ed425`:

- Deployed a patched Magnum image built from this PR head in an OVS-based Atmosphere environment.
- Rechecked the previously reported Tempest-style guest floating-IP SSH path with a CirrOS tenant VM. Ping and SSH through the floating IP both passed; the earlier timeout did not reproduce.
- Created a fresh Magnum cluster from the patched image. Result: `CREATE_COMPLETE`, `HEALTHY`, one control-plane node Ready, one worker node Ready.
- Cinder CSI scenario passed: `block-rbd1` PVC bound, a pod mounted it, wrote/read `cinder-ok`, exited successfully, and the backing Cinder volume detached cleanly.
- Kubernetes `Service type=LoadBalancer` scenario passed: the service received an external address, the backing Octavia load balancer was `ACTIVE`/`ONLINE`, and an HTTP probe returned `mcapi-ovs-lb-ok`.
- Missing MachineDeployment nodegroup delete scenario passed: created a non-default nodegroup, manually deleted its management-cluster MachineDeployment, then `openstack coe nodegroup delete` completed and the nodegroup returned HTTP 404.
- Invalid fixed network/subnet negative checks passed: both failed immediately with HTTP 400.
- Manila CSI was exercised and failed in the environment backend path, not in the Magnum SDK migration path. The PVC stayed Pending because Manila put the share in `error`: it could not find or allocate a share server on the supplied share network.
- Cleanup completed for the successful test cluster, workload namespaces, service load balancer, Cinder PVC volume, failed Manila share, test cluster template, router, subnet, and network. Earlier pre-fix failed cluster records also accepted delete after the fixed image was deployed.

Local validation for the final head:

- `uv run black --check magnum_cluster_api tests`
- `uv run isort --check-only magnum_cluster_api tests`
- `env -u KUBECONFIG uv run pytest magnum_cluster_api/tests/unit/ -q` -> `203 passed, 16 warnings`

PR #1038, opened 2026-05-21
fix: add shared sdk connection config

ricolin:
## Summary

- add an explicit `openstack_client` config group for the shared openstacksdk connection
- keep PR 1031 service-client endpoint and region options as compatibility fallback
- document the new config path and cover precedence with unit tests

## Testing

- `uv run pytest magnum_cluster_api/tests/unit/test_clients.py -q`
- `uv run black --check magnum_cluster_api tests`
- `uv run isort --check-only magnum_cluster_api tests`
- `KUBECONFIG=<temp> uv run pytest magnum_cluster_api/tests/unit/ -q` (`207 passed, 15 warnings`)
- `pre-commit run --all-files` with Python 3.10 hook env override (default local `python3` is 3.8 and cannot install Black 25.1.0)

## Live Validation

Pending on OVS AIO and OVN AIO.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+* add explicit shared OpenStack SDK connection configuration, with existing
+  service client options retained as compatibility fallback
+
 ## [0.36.6](https://github.com/vexxhost/magnum-cluster-api/compare/v0.36.5...v0.36.6) (2026-04-21)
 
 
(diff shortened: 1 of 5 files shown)


mnaser:
@ricolin can we re-use the existing configuration, as in, use the existing magnum config for keystone.. use the existing config for neutron, etc..

also, i think we should maybe make sure that we create a client using the token we have from the user context, so for example, when we call create_cluster, we should use the credentials from the API call inside the context

thinking about this, I feel like we should _never_ ever have to use the OpenStack credentials inside the Magnum configuration and always grab it from the context so we can act 'on behalf of' the OpenStack user to make sure proper RBAC is in iplace

I don't think the driver needs to do any admin level ops I think?

PR #1039, opened 2026-05-21
fix: use openstacksdk for application credentials

mnaser:
## Summary
- build an openstacksdk connection from Magnum's request context using the same session/auth split used by Nova and Heat
- use the SDK identity proxy for application credential create/find/delete and service discovery
- add direct openstacksdk dependency and unit coverage for context-derived auth

## Validation
- pre-commit run --all-files
- uv run pytest magnum_cluster_api/tests/unit/test_clients.py magnum_cluster_api/tests/unit/test_driver.py::TestDriver::test_create_cluster


ricolin:
@mnaser  We still need fixes from https://github.com/vexxhost/magnum-cluster-api/pull/1040 before this CI fix is ready

mnaser:
fixed in https://github.com/vexxhost/magnum-cluster-api/pull/1071

PR #1048, opened 2026-05-26
fix: validate server group member quota before provisioning

larainema:
## Summary
- add a Nova limits preflight for the server_group_members quota
- fail cluster, node group, and scale-up requests before creating CAPI/CAPO resources when the requested server group size is too large
- include autoscaler max_node_count in the validation and return an actionable Magnum error

## Testing
- /Users/dong/dongdev/VEXXHOST/magnum-cluster-api/.tox/unit/bin/python -m pytest magnum_cluster_api/tests/unit/test_utils.py -k 'server_group_members_quota' magnum_cluster_api/tests/unit/test_driver.py -k 'quota'
- uvx pre-commit run --files magnum_cluster_api/driver.py magnum_cluster_api/utils.py magnum_cluster_api/exceptions.py magnum_cluster_api/tests/conftest.py magnum_cluster_api/tests/unit/test_utils.py magnum_cluster_api/tests/unit/test_driver.py

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -69,9 +69,10 @@ class BaseDriver(driver.Driver):
         cluster.stack_id = utils.generate_cluster_api_name(self.k8s_api)
         cluster.save()
 
-        self.rust_driver.create_cluster(cluster)
-
         utils.validate_cluster(context, cluster)
+        utils.validate_cluster_server_group_members_quota(context, cluster)
+
+        self.rust_driver.create_cluster(cluster)
 
         return self._create_cluster(context, cluster)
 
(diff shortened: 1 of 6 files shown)


mnaser (changes requested):
I believe there was work upstream to enable a driver method that did pre-flight validation, can you check if that has landed, otherwise maybe we can push to get that to land and put all that logic there?

larainema:
> I believe there was work upstream to enable a driver method that did pre-flight validation, can you check if that has landed, otherwise maybe we can push to get that to land and put all that logic there?

I checked upstream Magnum master and Gerrit, and I do not see the generic driver pre-flight validation hook landed yet, I created two patches related this.

Relevant upstream Magnum reviews:

- https://review.opendev.org/c/openstack/magnum/+/994525
- https://review.opendev.org/c/openstack/magnum/+/994529

PR #1040, opened 2026-05-21
fix: keep PR 1039 scoped

ricolin:
## Summary

This is the cleanup layer on top of PR #1071. The workflow is now:

```text
main <- PR #1071 <- PR #1040
```

PR #1039 is no longer part of this stack.

## Fix

- Add `openstacksdk` as an explicit runtime dependency because PR #1071 imports SDK modules directly.
- Make SDK Cinder default volume type lookup tolerate `404`/`406` from `/types/default` by falling back to the first listed volume type.
- Delete SDK Octavia load balancers through SDK delete paths instead of passing SDK objects into Magnum's legacy Octavia helper.
- Wait for SDK load balancer deletion through SDK list calls.
- Parse `cinder_csi_enabled` and `manila_csi_enabled` as OpenStack string labels instead of native booleans.
- Update the default OCCM provider test expectation to `amphorav2`.

## Scope

This PR intentionally does not include the label fallback or missing CAPI `status.conditions` fixes. Those are owned by PR #1016.

## Validation

```bash
uv run --python 3.10 pytest \
  magnum_cluster_api/tests/unit/test_clients.py \
  magnum_cluster_api/tests/unit/test_utils.py -q
# 37 passed

cargo test
# 143 passed; doctest ignored

git diff --check
# passed
```

Full Python unit suite status:

```text
uv run --python 3.10 pytest magnum_cluster_api/tests/unit -q
193 passed, 8 failed
```

The 8 failures are the pre-existing `update_nodegroup` response-mock failures unrelated to this cleanup diff.

Signed-off-by: Rico Lin <rlin@vexxhost.com>

PR #1013, opened 2026-05-06
Support config profiles

ricolin:
## Usage

Operators define reusable profiles in the management Kubernetes cluster using `ConfigMap/mcapi-config-profiles`, normally in the `magnum-system` namespace. Users select those profiles through cluster template labels.

Supported profile fields:

- `kubeletConfig`
- `files`
- `preKubeadmCommands`
- `postKubeadmCommands`

Users select predefined profiles with cluster template labels:

| Label | Example | Effect |
|---|---|---|
| `config_profile` | `profile-standard` | Selects the cluster-wide default profile rendered as the `configProfile` ClusterClass topology variable. |
| `nodegroup_config_profile_set` | `profile-bm-gpu-layout` | Selects an operator-defined nodegroup layout that renders MachineDeployment-level `configProfile` overrides. |

Passing `config_profile` or `nodegroup_config_profile_set` directly through `openstack coe cluster create --labels` is rejected unless the value exactly matches the selected cluster template.

## Updates

This PR now also includes the Issue 42 live-replay fix:

- `fix(pr1013): keep worker pre-kubeadm list initialized`

That fix keeps worker `preKubeadmCommands` initialized for the containerd reload/restart commands and preserves them when profile pre-kubeadm commands initialize the same list. It was moved here from PR B because the failure belongs to the config-profile/pre-kubeadm composition path owned by PR #1013.

## Validation

```bash
cargo test containerd_config
cargo test kubelet_config
cargo test test_convert_values_to_cluster_topology_variables
uv run pytest magnum_cluster_api/tests/unit/test_utils.py -q -k 'ConfigProfile or config_profile or kubelet_config'
uv run ruff check magnum_cluster_api/tests/unit/test_utils.py
```

Latest scoped validation after moving Issue 42 into this PR:

- `cargo test containerd_config` -> 2 passed
- `cargo test kubelet_config` -> 8 passed
- `cargo test test_convert_values_to_cluster_topology_variables` -> 1 passed
- `uv run pytest ... -k 'ConfigProfile or config_profile or kubelet_config'` -> 35 passed, 16 deselected, 8 warnings
- `uv run ruff check magnum_cluster_api/tests/unit/test_utils.py` -> passed

Signed-off-by: Rico Lin <rico@vexxhost.com>

PR #1051, opened 2026-05-27
fix(labels): restore template labels for sparse cluster labels

ricolin:
## Purpose

This draft PR now keeps only the sparse cluster-template label fallback that is still not covered by PR #1040 plus PR #1013.

Magnum `cluster create --labels` replaces, rather than merges, cluster-template labels. If a user supplies any sparse cluster labels, the stored cluster row can miss template keys such as `kube_tag`. The Rust label extractor and monitor then fail with `KeyError: 'kube_tag'`, and delete/status paths can also trip over the same sparse label dict.

## Scope Kept

- Merge missing `cluster.cluster_template.labels` into `cluster.labels` before Rust create/delete paths, without overriding user-provided labels.
- Keep the related unit coverage and remove the old internal codename from the test docstring.

## Scope Dropped After Retest

These were removed from this draft because PR #1040 already covers them when combined with PR #1013:

- OpenStackSDK application credential/service/volume/flavor compatibility.
- OpenStackSDK server group compatibility.
- Rust string boolean label parsing.
- hardware disk bus `None` normalization.
- nodegroup MachineSet-list test fixture adjustment.
- eventlet/tpool avoidance changes, with no separate failure reproduced in the combined retest.
- default amphorav2 provider test expectation.

## Validation

```bash
uvx --python 3.12 pre-commit run --all-files
# passed

uv run --python 3.10 pytest \
  magnum_cluster_api/tests/unit/test_utils.py \
  -q -k 'FillMissingLabelsFromTemplate'
# 4 passed, 16 deselected, 7 warnings
```

Live AIO retest with a temporary combined PR #1040 + PR #1013 image:

- Built and rolled the combined image to the OVN AIO Magnum workloads.
- Normal config-profile template create was accepted and persisted template-derived labels including `kube_tag`.
- Sparse-label create with an extra non-profile label was accepted but persisted only the sparse labels plus `config_profile`; the conductor then logged `KeyError: 'kube_tag'` while extracting `Cluster.labels`.
- Config-profile override create failed as expected with the PR #1013 validation message.
- The OVS AIO was quota-blocked before create-path validation; fallback AIO hosts were unreachable from this session.

Signed-off-by: Rico Lin <rico@vexxhost.com>

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -70,6 +70,11 @@ class BaseDriver(driver.Driver):
         cluster.stack_id = utils.generate_cluster_api_name(self.k8s_api)
         cluster.save()
 
+        # See note in delete_cluster: Magnum REPLACES template labels with the
+        # cluster's own --labels list.  Merge any missing keys back in so the
+        # Rust driver and downstream consumers see a complete label dict.
+        utils.fill_missing_labels_from_template(cluster)
+
         self.rust_driver.create_cluster(cluster)
 
         utils.validate_cluster(context, cluster)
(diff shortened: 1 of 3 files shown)


ricolin:
Closing this draft as superseded by PR #1016. PR #1016 already carries the same sparse cluster-template label fallback, and I moved the live retest evidence there with an additional persistence fix so filled labels are saved for later monitor/status reads.

ricolin:
Superseded by #1016.

PR #1055, opened 2026-05-28
Fix conductor Kubernetes lock client usage

ricolin:
## Summary

- load Kubernetes authentication before creating Sherlock cluster locks
- pass an authenticated CoordinationV1 client into Sherlock and normalize bearer tokens for Lease requests
- call the Rust mCAPI Driver and KubeClient directly from Magnum conductor instead of wrapping PyO3 objects with eventlet tpool proxies
- normalize both `authorization` and generated-client `BearerToken` auth keys so Lease calls do not run as `system:anonymous`
- tolerate early CAPI Cluster objects that do not have `status.conditions` yet, so Magnum health/status refresh retries instead of crashing with `KeyError`
- isolate direct Rust/Kubernetes client constructors in unit tests so this PR does not depend on a local kubeconfig or live Kubernetes API

## Validation

- `uv run pytest magnum_cluster_api/tests/unit/test_sync.py magnum_cluster_api/tests/unit/test_driver.py::TestDriver::test_update_cluster_status_handles_missing_capi_conditions 'magnum_cluster_api/tests/unit/test_driver.py::test_base_driver_uses_direct_rust_clients' 'magnum_cluster_api/tests/unit/test_driver.py::TestDriver::test_create_cluster' -q` -> 15 passed, 16 warnings
- `uv run ruff check magnum_cluster_api/sync.py magnum_cluster_api/driver.py magnum_cluster_api/tests/unit/test_sync.py magnum_cluster_api/tests/unit/test_driver.py magnum_cluster_api/tests/conftest.py` -> passed

## Live Replay Notes

The OVN PyTorch BM replay on aio env hit two PR-1055-owned runtime issues after the initial direct-client changes:

- Sherlock Lease requests were still anonymous with the generated Kubernetes client because this client version reads `Configuration.api_key["BearerToken"]` instead of only `authorization`.
- Magnum health refresh saw an early CAPI Cluster object without `status.conditions` and crashed the periodic status path with `KeyError`, leaving stale health even after CAPI became Ready.

Both fixes were validated in the replay image before being moved here.

### DIFF
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -218,7 +218,8 @@ class BaseDriver(driver.Driver):
 
             capi_cluster.reload()
             status_map = {
-                c["type"]: c["status"] for c in capi_cluster.obj["status"]["conditions"]
+                c["type"]: c["status"]
+                for c in capi_cluster.obj.get("status", {}).get("conditions", [])
             }
 
             for condition in ("ControlPlaneReady", "InfrastructureReady", "Ready"):
--- a/magnum_cluster_api/driver.py
+++ b/magnum_cluster_api/driver.py
@@ -15,7 +15,6 @@
 
 from __future__ import annotations
 
-from eventlet import tpool  # type: ignore
 from heatclient import exc  # type: ignore
 from magnum import objects as magnum_objects  # type: ignore
 from magnum.common import exception as magnum_exception  # type: ignore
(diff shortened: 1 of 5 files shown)


mnaser:
@ricolin what's the point of this pr?

we had to enable the tpool because it was breaking and making the whole process stuck.. what are we solving here?

PR #1066, opened 2026-06-12
Update CAPI for Kubernetes v1.36 kubeadm feature gate compatibility

### LINKED ISSUE #1065: Kubernetes v1.36 cluster creation fails due to removed kubeadm feature gate
## Summary

Cluster creation with `kube_tag=v1.36.1` fails because CAPI generates kubeadm config with:

```yaml
featureGates:
  ControlPlaneKubeletLocalMode: true
````

In Kubernetes v1.36, `ControlPlaneKubeletLocalMode` was removed, so `kubeadm init` fails.

## Error

Inside the control plane VM:

```bash
sudo kubeadm init --config /run/kubeadm/kubeadm.yaml
```

fails with:

```text
error: featureGates: Invalid value: {"ControlPlaneKubeletLocalMode":true}: ControlPlaneKubeletLocalMode is not a valid feature name.
```

## Impact

Because `kubeadm init` fails:

```text
kube-apiserver never starts
port 6443 remains unreachable
CAPI cannot connect to the control plane
worker node cannot join
cluster stays in CREATE_IN_PROGRESS
```

## Root cause

Our current CAPI version is not compatible with Kubernetes v1.36. It still injects `ControlPlaneKubeletLocalMode=true`, but Kubernetes v1.36 no longer accepts that feature gate.

Upstream CAPI fix:

[https://github.com/kubernetes-sigs/cluster-api/pull/13177](https://github.com/kubernetes-sigs/cluster-api/pull/13177)

## Proposed fix

Update CAPI to a version that includes the upstream fix.

Temporary workaround: use a lower Kubernetes version, such as `v1.31.x` or `v1.32.x`, for current testing.

okozachenko1203:
Fixes #1065


mnaser (changes requested):
Let's use 1.13.2 and 0.14.4 to match Atmosphere.

PR #1068, opened 2026-06-16
chore(zuul): bump hydrophone Kubernetes versions

ricolin:
## Summary
- bump Hydrophone Zuul jobs to Kubernetes v1.33.12, v1.34.8, and v1.35.5
- add the v1.36.1 Hydrophone calico/cilium jobs to check and gate

## Testing
- uvx --python 3.12 pre-commit run --all-files


mnaser:
We still have `latest` in there so the next time we cut a release of capo-image-elements, it will break.  I will create a follow up PR that takes care of this and updates the script.

PR #1073, opened 2026-06-17
fix(clients): finish SDK compatibility cleanup

ricolin:
## Summary

This is the cleanup layer on top of PR #1071. It replaces closed PR #1040,
which GitHub would not allow reopening after its deleted base/head branches were
recreated.

The workflow is now:

```text
main <- PR #1071 <- this PR
```

PR #1039 is no longer part of this stack.

## Fix

- Add `openstacksdk` as an explicit runtime dependency because PR #1071 imports SDK modules directly.
- Make SDK Cinder default volume type lookup tolerate `404`/`406` from `/types/default` by falling back to the first listed volume type.
- Delete SDK Octavia load balancers through SDK delete paths instead of passing SDK objects into Magnum's legacy Octavia helper.
- Wait for SDK load balancer deletion through SDK list calls.
- Parse `cinder_csi_enabled` and `manila_csi_enabled` as OpenStack string labels instead of native booleans.
- Update the default OCCM provider test expectation to `amphorav2`.

## Scope

This PR intentionally does not include the label fallback or missing CAPI `status.conditions` fixes. Those are owned by PR #1016.

## Validation

```bash
uv run --python 3.10 pytest \
  magnum_cluster_api/tests/unit/test_clients.py \
  magnum_cluster_api/tests/unit/test_utils.py -q
# 37 passed

cargo test
# 143 passed; doctest ignored

git diff --check
# passed
```

Full Python unit suite status:

```text
uv run --python 3.10 pytest magnum_cluster_api/tests/unit -q
193 passed, 8 failed
```

The 8 failures are the pre-existing `update_nodegroup` response-mock failures unrelated to this cleanup diff.

Signed-off-by: Rico Lin <rlin@vexxhost.com>


ricolin:
Folded into PR #1071 as four fresh commits with detailed rationale in each commit message. The resulting PR #1071 tree matches this PR content, so this stacked draft is no longer needed.

PR #1074, opened 2026-06-17
fix(addons): parse CSI enable labels from strings

ricolin:
## Summary
- Model `cinder_csi_enabled` and `manila_csi_enabled` as OpenStack string labels instead of native booleans.
- Add helper methods that treat only case-insensitive `"true"` as enabled.
- Route the Cinder and Manila CSI addon `enabled()` methods through those helpers.

## Why
Magnum labels are delivered to the Rust driver as OpenStack label values, which are strings even when the label represents a boolean. The CSI enable labels were modeled as `bool`, so values such as `"true"` and `"false"` depended on Python extraction accepting native booleans instead of the string values users set on cluster templates or clusters.

Keeping Cinder and Manila together isolates this from PR #1071 because both addons share the same enable-label bug and both implementations need to return a bool from `enabled()`.

## Testing
- `cargo test` passed: 145 tests passed, doctest ignored.
- `git diff --check origin/main..HEAD` passed.

### DIFF
--- a/src/addons/cinder_csi.rs
+++ b/src/addons/cinder_csi.rs
@@ -201,7 +201,7 @@ impl ClusterAddon for Addon {
     }
 
     fn enabled(&self) -> bool {
-        self.cluster.labels.cinder_csi_enabled
+        self.cluster.labels.is_cinder_csi_enabled()
     }
 
     fn secret_name(&self) -> Result<String, ClusterError> {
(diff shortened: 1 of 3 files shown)


mnaser:
@ricolin isnt this essentially the change in https://github.com/vexxhost/magnum-cluster-api/pull/1074 ? maybe we can land that and help get a new contributor? :)

PR #1084, opened 2026-06-18
fix: support Python 3.14 registry loader

ricolin:
Update native Python dependencies in the lockfile so the registry loader can run on Alpine 3.24 with the default CPython 3.14 interpreter. This replaces the earlier temporary Python 3.13 cap with dependency versions that provide CPython 3.14 musllinux wheels.

Validation:
- uv run black --check magnum_cluster_api/tests/unit/test_resources.py
- uv run pytest magnum_cluster_api/tests/unit/test_resources.py
- git diff --check
docker build --target registry-loader --progress=plain --output type=cacheonly .


ricolin:
recheck

mnaser:
I think we're going to simply move away from this registry loader since it's not something we support formally anymore.

mnaser:
Superseeded by https://github.com/vexxhost/magnum-cluster-api/pull/1098

PR #1131, opened 2026-07-25
fix: make legacy ClusterClass repair explicit

larainema:
> Stacked on #1121. This follow-up changes the repair introduced there from
> automatic reconciliation to an explicit administrator operation.

## What changed from #1121

- Remove automatic legacy ClusterClass repair from driver startup and normal ClusterClass application.
- Add an explicit `magnum-cluster-api-repair-legacy-cluster-class` administrator CLI for the scoped repair implemented by #1121.
- Require interactive confirmation by default and support `--yes` for planned automation.
- Return a clear message when the class is already repaired or is outside the legacy scope.
- Document inventory, backup, repair, and rollout-monitoring steps.

## Why

As described in #1121, Clusters remain pinned to the ClusterClass recorded in `spec.topology.class`. Some ClusterClasses generated before `v0.36.1` render empty `files[].content` values when no proxy is configured. Newer Cluster API validation rejects those values, which can leave topology reconciliation blocked after an upgrade.

Changing a shared ClusterClass can also update bootstrap templates for every Cluster using it and trigger Machine rollouts. For that reason, this PR makes the repair an explicit, one-class-at-a-time administrator action instead of applying it automatically during conductor startup.

## Operator workflow

```bash
magnum-cluster-api-repair-legacy-cluster-class \
  --yes \
  magnum-v0.34.2
```

Before running the command:

1. Inventory every Cluster using the target ClusterClass.
2. Pause affected Clusters and stop concurrent upgrades.
3. Back up the ClusterClass.
4. Confirm sufficient Nova capacity for control-plane and worker surge.
5. Repair the shared class, then unpause and monitor Clusters one at a time.

### Unit and source checks

- `uv run --no-sync pytest magnum_cluster_api/tests/unit/test_repair_legacy_cluster_class.py magnum_cluster_api/tests/unit/test_driver.py::test_driver_init_does_not_apply_cluster_class magnum_cluster_api/tests/unit/test_driver.py::TestDriver::test_create_cluster -q`
  - `7 passed`
- `git diff --check origin/codex/fix-legacy-clusterclass-proxy-content...HEAD`
- The final source tree was successfully built into a wheel and installed in the AIO environment.

The local Mac currently has Rust `1.90.0`, while this repository requires Rust `1.95`, so the final local `cargo test` invocation could not start. Rust tests should run in CI with the repository-supported toolchain.

### Active legacy AIO test

The test used a healthy, active one-control-plane/one-worker Cluster on a deliberately vulnerable `magnum-v0.36.0-aio-active` ClusterClass. Both Machines and workload Nodes were Running before executing the exact repair CLI.

| Role | State | Machine UID | Nova server ID |
| --- | --- | --- | --- |
| Control plane | Before repair | `2f815010-84ff-487c-bcd1-a90f6b4d3e4e` | `a11164e3-f54e-450e-929a-c80ecd2e26ea` |
| Control plane | After repair | `a5c6a5fc-96a0-4fd4-9db4-ce500238d365` | `686c6709-e5e5-4263-86cd-14be8075bf16` |
| Worker | Before repair | `acb127c4-b7cf-422a-aa9e-4748e97a0300` | `f4d9f563-4085-4134-a3ef-d47f843bb6a9` |
| Worker | After repair | `12df708b-7c49-4bcb-bf04-dba899ffbf36` | `4fccacfc-6e3d-4d13-89ac-e08ead444ea6` |

Observed result:

- The CLI reported `Repaired ClusterClass magnum-system/magnum-v0.36.0-aio-active.`
- All four affected proxy templates matched the repaired content.
- KubeadmControlPlane and MachineDeployment each surged from one to two Machines.
- New Machines became Running before their legacy counterparts were removed.
- Both original Nova server IDs were deleted.
- Final KubeadmControlPlane and MachineDeployment status was `1/1 Ready`, `1 updated`, and `0 unavailable`.
- Both workload Nodes were Ready and all 22 `kube-system` pods were Running.
- Magnum remained `CREATE_COMPLETE / HEALTHY`.
- One workload API poll returned no data during the single-control-plane handoff and recovered on the next poll. A maintenance window is therefore recommended for single-control-plane clusters.
- The disposable AIO resources were removed after validation.

## Merge, release, and backport

Merge this stacked change into the #1121 branch so the main PR contains the explicit workflow before it lands. Then identify the exact magnum-cluster-api release branch used by the target deployment before backporting. Do not infer the branch from deployment dates. The deployed package should include this CLI before operators repair affected legacy ClusterClasses.

### DIFF
--- a/docs/admin/troubleshooting.md
+++ b/docs/admin/troubleshooting.md
@@ -174,7 +174,12 @@ configured. Newer Cluster API validation rejects the empty value and leaves
 clusters using the class with `TopologyReconciled=False`.
 
 Do not repair every legacy ClusterClass automatically. Changing a ClusterClass
-causes all Clusters using it to reconcile and may trigger Machine rollouts.
+causes all Clusters using it to reconcile and triggers control plane and worker
+Machine rollouts. Schedule a maintenance window for every affected Cluster and
+confirm sufficient infrastructure capacity for replacement Machines. A Cluster
+with a single control plane Machine can experience a brief Kubernetes API
+interruption during the replacement.
+
 First, identify the Clusters using the affected class and stop any cluster
 upgrades:
 
@@ -183,6 +188,14 @@ $ export CLUSTER_CLASS=magnum-v0.34.2
 $ kubectl -n magnum-system get clusters \
     -o custom-columns=NAME:.metadata.name,CLASS:.spec.topology.class \
     | grep "$CLUSTER_CLASS"
+```
+
+Pause every Cluster returned by the inventory before repairing the shared
+ClusterClass:
+
+```
+$ kubectl -n magnum-system annotate cluster <cluster-name> \
+    cluster.x-k8s.io/paused=true --overwrite
 ```
 
 Back up the ClusterClass before changing it:
@@ -199,10 +212,13 @@ Repair only that ClusterClass. The command asks for confirmation unless
 $ magnum-cluster-api-repair-legacy-cluster-class "$CLUSTER_CLASS"
 ```
 
-Monitor topology reconciliation and Machine identities after applying the
-repair:
+Unpause and monitor Clusters one at a time. Wait for each Cluster's topology,
+control plane, worker Machines, and workload Nodes to become healthy before
+unpausing the next Cluster:
 
 ```
+$ kubectl -n magnum-system annotate cluster <cluster-name> \
+    cluster.x-k8s.io/paused- --overwrite
 $ kubectl -n magnum-system get clusters \
     -o custom-columns=NAME:.metadata.name,CLASS:.spec.topology.class,TOPOLOGY:.status.conditions[?(@.type==\"TopologyReconciled\")].status
 $ kubectl -n magnum-system get machines \
--- a/docs/admin/troubleshooting.md
+++ b/docs/admin/troubleshooting.md
@@ -165,3 +165,46 @@ Once the cluster is gone, you can clean up the project:
 $ unset OS_PROJECT_ID
 $ openstack project delete $CURRENT_PROJECT_ID
 ```
+
+## Legacy ClusterClass fails topology reconciliation after an upgrade
+
+ClusterClasses created by versions of the driver older than `v0.36.1` can
+contain proxy file patches whose rendered `content` is empty when no proxy is
+configured. Newer Cluster API validation rejects the empty value and leaves
+clusters using the class with `TopologyReconciled=False`.
+
+Do not repair every legacy ClusterClass automatically. Changing a ClusterClass
+causes all Clusters using it to reconcile and may trigger Machine rollouts.
+First, identify the Clusters using the affected class and stop any cluster
+upgrades:
+
+```
+$ export CLUSTER_CLASS=magnum-v0.34.2
+$ kubectl -n magnum-system get clusters \
+    -o custom-columns=NAME:.metadata.name,CLASS:.spec.topology.class \
+    | grep "$CLUSTER_CLASS"
+```
+
+Back up the ClusterClass before changing it:
+
+```
+$ kubectl -n magnum-system get clusterclass "$CLUSTER_CLASS" -o yaml \
+    > "$CLUSTER_CLASS.before-repair.yaml"
+```
+
+Repair only that ClusterClass. The command asks for confirmation unless
+`--yes` is supplied:
+
+```
+$ magnum-cluster-api-repair-legacy-cluster-class "$CLUSTER_CLASS"
+```
+
+Monitor topology reconciliation and Machine identities after applying the
+repair:
+
+```
+$ kubectl -n magnum-system get clusters \
+    -o custom-columns=NAME:.metadata.name,CLASS:.spec.topology.class,TOPOLOGY:.status.conditions[?(@.type==\"TopologyReconciled\")].status
+$ kubectl -n magnum-system get machines \
+    -o custom-columns=NAME:.metadata.name,UID:.metadata.uid,CLUSTER:.metadata.labels.cluster\\.x-k8s\\.io/cluster-name
+```
--- a/magnum_cluster_api/cmd/repair_legacy_cluster_class.py
+++ b/magnum_cluster_api/cmd/repair_legacy_cluster_class.py
@@ -1,0 +1,45 @@
+# Copyright (c) 2026 VEXXHOST, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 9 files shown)


src/legacy_cluster_class.rs
```diff
@@ -19,30 +19,32 @@ const SYSTEMD_PROXY_PLACEHOLDER_TEMPLATE: &str =
 pub async fn repair_legacy_proxy_file_patches(
     client: Client,
     namespace: &str,
-) -> Result<usize, kubernetes::Error> {
+    cluster_class_name: &str,
+) -> Result<bool, kubernetes::Error> {
```
okozachenko1203: just true/false response is not clear. I'd define repair result as enum to distinguish the cases between already-fixed or out-of-scope.

magnum_cluster_api/cmd/repair_legacy_cluster_class.py
```diff
@@ -0,0 +1,45 @@
...
+
+    repaired = magnum_cluster_api.Driver(namespace).repair_legacy_cluster_class(
+        cluster_class_name
+    )
+    if repaired:
+        click.echo(f"Repaired ClusterClass {namespace}/{cluster_class_name}.")
+    else:
+        click.echo(
+            f"No repair was needed for ClusterClass {namespace}/{cluster_class_name}."
+        )
```
okozachenko1203: here, once we define repair result enum in rust code, need to add elif condition also

docs/admin/troubleshooting.md
```diff
@@ -165,3 +165,46 @@ Once the cluster is gone, you can clean up the project:
...
+ClusterClasses created by versions of the driver older than `v0.36.1` can
+contain proxy file patches whose rendered `content` is empty when no proxy is
+configured. Newer Cluster API validation rejects the empty value and leaves
+clusters using the class with `TopologyReconciled=False`.
+
+Do not repair every legacy ClusterClass automatically. Changing a ClusterClass
+causes all Clusters using it to reconcile and may trigger Machine rollouts.
+First, identify the Clusters using the affected class and stop any cluster
+upgrades:
+
```
okozachenko1203, line 185: need to explicitly mention that users need a maintenance window for single cp nodes

docs/admin/troubleshooting.md
```diff
@@ -165,3 +165,46 @@ Once the cluster is gone, you can clean up the project:
...
+```
+$ export CLUSTER_CLASS=magnum-v0.34.2
+$ kubectl -n magnum-system get clusters \
+    -o custom-columns=NAME:.metadata.name,CLASS:.spec.topology.class \
+    | grep "$CLUSTER_CLASS"
```
okozachenko1203, line 190: I'd add reconcile pause step
```suggestion
$ export CLUSTER_CLASS=magnum-v0.34.2
$ kubectl -n magnum-system get clusters \
    -o custom-columns=NAME:.metadata.name,CLASS:.spec.topology.class \
    | grep "$CLUSTER_CLASS"
$ kubectl -n magnum-system annotate cluster <cluster-name> \
    cluster.x-k8s.io/paused=true --overwrite
```

docs/admin/troubleshooting.md
```diff
@@ -165,3 +165,46 @@ Once the cluster is gone, you can clean up the project:
...
+
+Repair only that ClusterClass. The command asks for confirmation unless
+`--yes` is supplied:
+
+```
+$ magnum-cluster-api-repair-legacy-cluster-class "$CLUSTER_CLASS"
+```
+
+Monitor topology reconciliation and Machine identities after applying the
+repair:
```
okozachenko1203:
```suggestion
Unpause and monitor Clusters one at a time. Watch topology reconciliation and
Machine identities after applying the repair:
```

docs/admin/troubleshooting.md
```diff
@@ -165,3 +165,46 @@ Once the cluster is gone, you can clean up the project:
...
+
+```
+$ kubectl -n magnum-system get clusters \
+    -o custom-columns=NAME:.metadata.name,CLASS:.spec.topology.class,TOPOLOGY:.status.conditions[?(@.type==\"TopologyReconciled\")].status
+$ kubectl -n magnum-system get machines \
+    -o custom-columns=NAME:.metadata.name,UID:.metadata.uid,CLUSTER:.metadata.labels.cluster\\.x-k8s\\.io/cluster-name
```
okozachenko1203, line 225:
```suggestion
$ kubectl -n magnum-system annotate cluster <cluster-name> \
    cluster.x-k8s.io/paused- --overwrite
$ kubectl -n magnum-system get clusters \
    -o custom-columns=NAME:.metadata.name,CLASS:.spec.topology.class,TOPOLOGY:.status.conditions[?(@.type==\"TopologyReconciled\")].status
$ kubectl -n magnum-system get machines \
    -o custom-columns=NAME:.metadata.name,UID:.metadata.uid,CLUSTER:.metadata.labels.cluster\\.x-k8s\\.io/cluster-name
```

okozachenko1203 (changes requested):

okozachenko1203:
@larainema @yaguangtang  let's re-evaluate this solution.
It's reasonable to make the patch triggerable rather than auto. This process requires manually executing the new patch command for each cluster class, pausing/unpausing coe clusters one by one, and finally triggering the cluster machine rollout. From an operational and HA perspective, this seems to offer zero difference from manually upgrading a COE cluster with a new cluster class. Therefore, the goal of this patch is unclear. And I'd let customers to upgrade COE clusters one by one.

larainema:
> Dong Ma (@larainema) Yaguang Tang (@yaguangtang) let's re-evaluate this solution. It's reasonable to make the patch triggerable rather than auto. This process requires manually executing the new patch command for each cluster class, pausing/unpausing coe clusters one by one, and finally triggering the cluster machine rollout. From an operational and HA perspective, this seems to offer zero difference from manually upgrading a COE cluster with a new cluster class. Therefore, the goal of this patch is unclear. And I'd let customers to upgrade COE clusters one by one.

Thanks for raising this. Based on the AIO testing, I agree that we should re-evaluate the purpose of this change.
The original intent of PR #1131 was to replace the automatic repair from PR #1121 with an explicit, narrowly scoped, and idempotent recovery command. This would allow operators to control when each legacy ClusterClass is modified instead of triggering unplanned reconciliation during driver startup.
However, the final AIO test confirmed that modifying the ClusterClass triggers reconciliation and replaces both control-plane and worker Machines. The Machine UIDs and Nova server IDs changed after the repair, so this is not an in-place correction and does not provide a no-rollout recovery path.
If an affected Cluster can be upgraded normally to a newer, valid ClusterClass, I agree that upgrading COE clusters one at a time provides the same rollout outcome with better isolation. It also avoids modifying a shared legacy ClusterClass used by multiple clusters.
The repair command would only provide additional value as a break-glass recovery path if the legacy validation failure prevents the normal COE upgrade from switching the Cluster to a newer ClusterClass.

PR #1145, opened 2026-08-13
fix(mcapi): allow external fixed networks for bare metal

ricolin:
## Summary

Allow a `server_type=bm` cluster to select an external Neutron provider network as its fixed network. Bare-metal control-plane and worker machines commonly attach directly to a physical L2 provider network, which Neutron correctly marks external.

- Attempt the existing non-external fixed-network lookup first.
- For bare-metal templates only, fall back to an external-network lookup.
- Keep the VM path unchanged: VM templates remain restricted to non-external fixed networks.
- Apply the same rule in validation and resource rendering.

## Why this is a production fix

The current VM-oriented validation rejects a valid bare-metal provider network before Cluster API Provider OpenStack can create infrastructure. Operators are then forced into an artificial tenant network or local code modifications, even though their physical machines and API endpoint must share the provider L2 segment.

This is intentionally not a general relaxation of Neutron validation. It is limited to the explicit bare-metal server type supplied by the prerequisite below.

## Dependency

Depends on #1014, which introduces `server_type=bm` and the bare-metal placement contract.

## Non-goals

This change does not create provider networks, change Neutron port-security policy, or alter VM cluster networking. Those remain operator and deployment responsibilities.

## Tests

- `pytest -q magnum_cluster_api/tests/unit/test_utils.py` — 38 passed
- Added coverage for external fixed-network resolution and validation for both bare-metal and VM templates.


ricolin:
### Failure mode addressed

Today `get_fixed_network_id()` resolves a fixed network with `external=False` unconditionally. A provider network attached directly to a physical bare-metal NIC is commonly marked `router:external=true`, so that lookup raises `FixedNetworkNotFound` before mCAPI can render the OpenStackCluster.

The result is a rejected cluster request, not a runtime networking failure. The fallback is deliberately reachable only after the standard lookup fails and only when the template explicitly selects `server_type=bm`; VM templates retain the existing safety boundary.

ricolin:
@codex[agent] review this PR

PR #1146, opened 2026-08-13
feat(mcapi): support lifecycle-safe clusters without Octavia

ricolin:
## Summary

Add a complete, opt-in control-plane path for environments that do not provide Octavia. The existing behavior remains the default.

With `octavia_enabled=false`, an operator may supply a managed fixed API address and enable kube-vip:

- CAPO renders no API load balancer or API floating IP.
- The fixed API address is propagated through ClusterClass and preserved across immutable-field reconciliation.
- kube-vip is rendered only for control-plane bootstrap, with an explicit interface and image.
- kube-vip uses the bootstrap `super-admin.conf` identity for leader election.
- OCCM retains its default cloud-node, node-lifecycle, and route controllers, while removing only the Octavia-backed `service` controller.
- Delete skips legacy Octavia cleanup only for clusters created with Octavia disabled.

## Why this is a production fix

A cluster without Octavia still requires a stable Kubernetes API endpoint and a healthy cloud controller manager. The previous implementation assumed an Octavia API load balancer, and therefore cannot create, reconcile, or delete this deployment model reliably.

The implementation is lifecycle-safe: it tracks whether the API address is operator-managed, preserves CAPO immutable fields during later reconciles, and keeps legacy behavior intact for existing clusters.

## Dependency

Depends on #1016. Magnum cluster labels can be sparse because cluster-level labels replace template labels; #1016 restores missing template defaults so creation and deletion receive a consistent configuration.

## Compatibility and scope

- `octavia_enabled` defaults to `true`; existing clusters retain Octavia behavior.
- `Service.type=LoadBalancer` is intentionally unavailable when Octavia is disabled.
- The fixed address, interface, and image are operator inputs; no environment-specific values are embedded.

## Tests

- Focused Python delete regression: 4 passed.
- Focused Rust tests: fixed-IP rendering (2), kube-vip rendering/gating (2), OCCM controller selection (1), and topology variables (1) all passed.
- The local full suites additionally contain existing Helm-render tests; those require the `helm` executable, which is not installed on this workstation. CI remains the full-suite authority.


ricolin:
### Failure modes addressed by the combined change

These paths must be shipped together because each fixes a different stage of the same no-Octavia lifecycle:

1. Without a managed fixed API address and kube-vip, disabling CAPO load balancing leaves no stable API endpoint.
2. Mounting the normal kubeadm admin kubeconfig into kube-vip is insufficient during initial bootstrap: its RBAC binding is created only after the API becomes available, while kube-vip needs the API to acquire its leader-election lease. This is a circular bootstrap failure. `super-admin.conf` exists for that early phase.
3. OCCM normally starts the `service` controller, which initializes Octavia-backed `Service.type=LoadBalancer` reconciliation. When Octavia is intentionally absent, this can prevent the shared OCCM process from becoming healthy, so ProviderIDs and node lifecycle reconciliation do not run.
4. CAPO treats the API fixed IP as immutable. Without ownership tracking and immutable-field preservation, a later reconcile with sparse labels can attempt to remove or alter the address.
5. A delete path that always performs Octavia cleanup fails a cluster that never created an Octavia resource.

The implementation keeps the default Octavia path unchanged and removes only the unsupported service-controller capability when the opt-in label is false.

ricolin:
@codex[agent] review this PR

